PR #3, opened 2014-03-14
Fix LP bug 1243827

adeuring:
If the text of a scalar node of the parsed tree starts with a digit or '+' or '-', the function resolve() removes all '_' from it and tries to convert the new string into a float or integer. When this fails, the variant without the '_' was returned as a string value, which is obviously wrong.

I also removed the loop where strings.Replace(in, "_", "", -1) was called with a with a simple call of replace(). This should have the same result.

The changes in this request do not explain how the workaround mentioned in the bug report worked: strings.Replace() just removes _all_ underscores -- this affects double underscores as well as single underscores. Also, I was not able to reproduce it, neither with the current versions of yaml and juju-core nor with the juju-core version from the "stable" PPA for precise.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -350,6 +350,20 @@ var unmarshalTests = []struct {
 			C inlineB `yaml:",inline"`
 		}{1, inlineB{2, inlineC{3}}},
 	},
+
+	// bug 1243827
+	{
+		"a: -b_c",
+		map[string]interface{}{"a": "-b_c"},
+	},
+	{
+		"a: +b_c",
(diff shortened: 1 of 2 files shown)


niemeyer:
LGTM, thank you.

niemeyer:
FWIW, the point of the loop was not because Replace would remove only a single entry. Note how it actually breaks out of the loop when it finds the first one. The point was to not call Replace in the first place, if it would do nothing. The change is probably making it slower. But Replace still avoids the allocation these days, if there's nothing to do, so hopefully not much slower.

PR #10, opened 2014-05-01
Cannot marshal/unmarshal time.Time

ewalk153:
This only answers half the problem: encoding. Test included as well.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -4,6 +4,7 @@ import (
 	"reflect"
 	"sort"
 	"strconv"
+	"time"
 )
 
 type encoder struct {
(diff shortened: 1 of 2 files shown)


niemeyer:
I'm happy to merge this. Can you please sign the contribution agreement:

```
http://www.canonical.com/contributor-license-agreement
```

It's a straightforward/non-draconian one.

ewalk153:
It should be properly filled out.

niemeyer:
Thanks!  Can you include the decoding side, now that the non-pointer bug is fixed?

ewalk153:
Yeah, I'll take another look.

niemeyer:
Ah, we have to look at something else too. The time.Time encoding and decoding format must match the standard yaml timestamp type: http://yaml.org/type/timestamp.html

ewalk153:
Fixing the formatting should be quick here.

The decoding looks to be a little more complicated. I'm not sure how to infer the type of the target field. Any suggestions?

niemeyer:
We can just try them all in sequence, starting by the format we marshal out so that the default encode/decode sequence hits on the first try.

ewalk153:
I'm getting stuck using reflection to assign the out value.

As a `time.Time` struct has only exported fields, I'm not sure out to assign to it.

http://play.golang.org/p/Z-pI2fm_Fa

niemeyer:
We need time handling, but we need both sides of it. I'm closing this for lack of activity. Please feel free to push the complete change once you're ready.

Issue #22, opened 2014-08-07
return ordered mappings?


termie:
in applications the order of things in yaml and json tend to matter, but it appears we're using a map[interface{}]interface{} as the underlying datastructure for what are conceptually ordered maps.

Even after implementing an OrderedMap I can't populate its keys in the order specified in the yaml because it is returned with go's officially randomly sorted maps. Most other parsers in other languages i've tried return an ordered mapping for something like this:

```
- s3sync:
    key_id: $AWS_ACCESS_KEY_ID
    key_secret: $AWS_SECRET_ACCESS_KEY
    bucket_url: $AWS_BUCKET_URL
    source_dir: build/
```

becomes:

```
{"sync": {"key_id": ...., "key_secret": ...}}
```

mattes:
run into this as well today

mattes:
```
// Example usage:
//    type T map[yaml.IndexMap0]interface{}
//    t := T{}
//    yaml.Unmarshal(data, &t)
```

niemeyer:
Most applications actually don't care about the ordering of maps in either JSON or YAML, and all parsers I know will also unmarshal into a first class hash map of the host language, which is pretty much always unordered, with the notable exception of JavaScript. For example, this is Python:

```
>>> import yaml
>>> yaml.load("{b: 2, a: 1}")
{'a': 1, 'b': 2}
```

The YAML [specification](http://yaml.org/spec/1.1/#id932806) is also pretty clear about what maps are:

> A mapping node is an unordered collection of key: value pairs.

With all that said, we can support ordered maps. I have documented in #30 the plan for that. Let's please follow up the conversation on that issue.

PR #27, opened 2014-09-07
Allow "ordered" mappings with custom StringIndex type

mattes:
Golang doesn't store the order of maps. So in a struct `map[string]string`  and YAML file like:

``` yaml
a: foo
b: bar
```

... there is no garantuee that `a` comes before `b`! Sometimes it might be useful though to know the order. My patch introduces a new type called `StringIndex`. It stores the index of the elements in maps according to their column, so that multiple `map[yaml.StringIndex]` are possible.

At the moment, I can't think of a better way of doing it. Please comment!

See also https://github.com/go-yaml/yaml/issues/22


mattes:
Ah this patch fixes `decode_test.go` and `encode_test.go` as well.

niemeyer:
A map for which you can never anticipate the key is not a very useful building block.

The proper way to support this would be with something like:

```
type MapSlice []MapItem

type MapItem struct {
        Key, Value interface{}
}
```

and then implement first-class marshalling and unmarshalling of maps onto it.

niemeyer:
Opened issue #30 to track the work. Closing the pull request as that's not the right direction.

PR #33, opened 2014-09-15
marshal top level nil correctly

rogpeppe:
This prevents yaml.Marshal(nil) from panicking.


niemeyer:
Also merged in v2 on b6b591a3.

PR #43, opened 2014-10-10
Bugfix: quote strings containing colon-space

riobard:
Unquoted strings containging ": " (colon-space) will confuse parsers due
to the ambiguity of inling mapping. See
http://www.yaml.org/YAML_for_ruby.html#indicators_in_strings for detail.


niemeyer:
This should be fixed, but it's not clear this is the right fix. Shouldn't it be quoted rather than turned into a literal? Also, can we please have a test for this?

niemeyer:
Closing for lack testing and lack of feedback. I cannot reproduce this issue, and have added a test at 71e7ede to ensure it continues to be the case. Please feel free to reopen or open an issue if you have a test case for it.

Issue #49, opened 2014-10-29
uint64 loses data when unmarshalling


liggitt:
Currently, any uint64 which overflows int64 gets unmarshalled as a float64 instead.

At best, this loses precision (since the float64 only has 52 bits to store the fraction in).

At worst, a bug in the boundary condition prevents the unmarshalling from happening at all:
`if resolved < 1<<64-1`

This should be `if resolved <= 1<<64-1`...

Test case:

```
package main

import (
    "fmt"
    "math"
    "strconv"

    "gopkg.in/yaml.v2"
)

func main() {
    fmt.Println("\nTesting loss of precision\n=======================================")
    test(12345678901234567890)

    fmt.Println("\nTesting uint64 max\n=======================================")
    test(math.MaxUint64)
}

type Test struct {
    X uint64
}

func test(x uint64) {
    m := Test{x}

    fmt.Printf("Original (base-2):      %s\n", strconv.FormatUint(m.X, 2))
    fmt.Printf("Original (base-10):     %s\n\n", strconv.FormatUint(m.X, 10))

    y, _ := yaml.Marshal(m)
    fmt.Printf("Marshalled YAML:     %s\n", string(y[:]))

    var m_yaml Test
    yaml.Unmarshal(y, &m_yaml)
    fmt.Printf("Unmarshalled (base-2):  %s\n", strconv.FormatUint(m_yaml.X, 2))
    fmt.Printf("Unmarshalled (base-10): %s\n\n", strconv.FormatUint(m_yaml.X, 10))
}
```

liggitt:
output of the test case:

```
$ go run yaml.go 

Testing loss of precision
=======================================
Original (base-2):      1010101101010100101010011000110011101011000111110000101011010010
Original (base-10):     12345678901234567890

Marshalled YAML:     x: 12345678901234567890

Unmarshalled (base-2):  1010101101010100101010011000110011101011000111110000100000000000
Unmarshalled (base-10): 12345678901234567168


Testing uint64 max
=======================================
Original (base-2):      1111111111111111111111111111111111111111111111111111111111111111
Original (base-10):     18446744073709551615

Marshalled YAML:     x: 18446744073709551615

Unmarshalled (base-2):  0
Unmarshalled (base-10): 0
```

PR #50, opened 2014-10-29
Add uint64 support to v1

liggitt:
Resolves https://github.com/go-yaml/yaml/issues/49 for v1

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -3,6 +3,7 @@ package yaml
 import (
 	"encoding/base64"
 	"fmt"
+	"math"
 	"reflect"
 	"strconv"
 	"time"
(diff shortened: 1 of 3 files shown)


niemeyer:
Thanks for the contribution. I'll provide comments inline.

Besides any comments, though, we need to apply such changes to v2 rather than v1. There's no good reason not to move on from v1.

liggitt:
@niemeyer ok, I'll open a PR against v2 if you want to wait a minute before commenting...

liggitt:
See https://github.com/go-yaml/yaml/pull/51

decode.go
```diff
@@ -356,8 +357,18 @@ func (d *decoder) scalar(n *node, out reflect.Value) (good bool) {
 				out.SetInt(resolved)
 				good = true
 			}
+		case uint:
```
niemeyer, line 360: It's not clear why there's special support for uint being added. The reason we handle int specially is because it's the default representation for integer literals, and is widely used by default for lengths, etc. What you are adding, though, is a special case that will come into play mainly when int64 overflows, in which case uint is irrelevant. We really want the 64 bits type to solve the problem.

liggitt, line 360: I'm fine with removing uint in lieu of uint64

PR #51, opened 2014-10-29
Add uint64 support to v2

### LINKED ISSUE #49: uint64 loses data when unmarshalling

liggitt:
Resolves #49 for v2

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -269,12 +269,12 @@ var unmarshalTests = []struct {
 
 	// int
 	{
-		"int_max: 2147483647", // math.MaxInt32
-		map[string]int{"int_max": 2147483647},
+		"int_max: 2147483647",
+		map[string]int{"int_max": math.MaxInt32},
 	},
 	{
-		"int_min: -2147483648", // math.MinInt32
-		map[string]int{"int_min": -2147483648},
+		"int_min: -2147483648",
+		map[string]int{"int_min": math.MinInt32},
 	},
 	{
 		"int_overflow: 9223372036854775808", // math.MaxInt64 + 1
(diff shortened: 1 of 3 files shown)


liggitt:
@niemeyer removed uint cases, left uint64 handling. Also inlined testcases as literal values to make it easier to follow, and added 64-bit support to the binary int decoding.

decode_test.go
```diff
@@ -267,6 +267,114 @@ var unmarshalTests = []struct {
 		map[string]uint64{},
 	},
 
+	// int
+	{
+		"int_max: 2147483647", // math.MaxInt32
```
niemeyer: Instead of having comments on all these entries, it might be better and more accurate to just use the actual value in the assertion. For example:

```
"int_max: 2147483647",
map[string]int{"int_max": math.MaxInt32},
```

niemeyer:
This looks good and I'm happy to merge it. Just see what you think of the suggestion above please.

decode_test.go
liggitt: I like it, will update

liggitt:
@niemeyer ready for merge

niemeyer:
Jordan, I got a bit ahead of myself here. I should have asked you to sign the contributor agreement first:

http://www.ubuntu.com/legal/contributors

It's a straightforward agreement, easy to sign online. Can you please have a look and see if it makes sense?  We need that for contributions.

liggitt:
Let me check, and I'll let you know as soon as I get the go ahead

niemeyer:
I'll need to back out the change if I don't get the signature this week. Sorry for the trouble, but this is required and I screwed up by not requesting beforehand.

niemeyer:
Actually, to avoid creating a larger problem for people that update now, I have backed out the change right now. Please just ping me once you get the signature in place, and I'll happily reapply.

liggitt:
No problem, I'll update as soon as I can

niemeyer:
Thanks, and sorry for the trouble.

niemeyer:
Any news there?

liggitt:
Haven't forgotten... still waiting to hear back, sorry

liggitt:
@niemeyer just got approval and signed the contributor agreement. Sorry for the delay.

liggitt:
@niemeyer you want me to open another PR or do you want to re-open/merge this one?

niemeyer:
Please leave it with me. I'll just reapply the change.

Thanks for handling the bureaucracy.

liggitt:
@niemeyer thanks... I'd already spawned #53, in case that makes it easier

PR #52, opened 2014-11-05
Fixes TypeError check in Unmarshal

BrianBland:
decoder.terrors can be empty but non-nil if the unmarshaler function is called multiple times in UnmarshalYAML.
This fixes the case where d.terrors is empty but has a non-zero capacity, which previously resulted in returning a TypeError with an empty list of error strings.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -647,6 +647,37 @@ func (s *S) TestUnmarshalerError(c *C) {
 	c.Assert(err, Equals, failingErr)
 }
 
+type sliceUnmarshaler []int
+
+func (su *sliceUnmarshaler) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var slice []int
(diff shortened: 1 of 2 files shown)


niemeyer:
Thanks for submitting this. Can we have a test covering the need for the change, please?

Also, would you mind to sign the contribution agreement if you haven't already done so?

```
https://www.ubuntu.com/legal/contributors
```

It's straightforward/non-draconian, and easy to sign online.

Thanks!

BrianBland:
Ok, I've added a test which fails before the patch and signed the agreement.

BrianBland:
Can I get an update on this PR? Thanks.

stevvooe:
@niemeyer Ping!

niemeyer:
Thanks, and sorry for the long delay on this.

stevvooe:
@niemeyer Thank you!

PR #57, opened 2014-11-25
Fix: Call UnmarshalYAML on literal empty string values

ghodss:
If you have the YAML `a: ""\n`, and the `a` field corresponds to a struct, UnmarshalYAML is not called on the struct. Instead, decoder.prepare(...) short-circuits and fails to parse with an error: `cannot unmarshal !!str``into <struct>`.

I've added an example at the end of decode_test.go, but a practical example exists in the [Kubernetes project struct util.IntOrString](https://github.com/GoogleCloudPlatform/kubernetes/blob/master/pkg/util/util.go#L63). With IntOrString, a YAML value turns into an IntOrString object which is created and filled in differently whether the value is an int or string in the YAML. So, `a: ""\n` and `a: 0\n` should both have UnmarshalYAML called on an IntOrString object, but `a: null\n` and `a: \n` both should have a zeroed out IntOrString object automatically created (or a nil pointer if the type is *IntOrString).

YAML v1 has the correct behavior since `a: ""\n` results in a call to UnmarshalYAML, and `a: null\n` and `a: \n` result in zeroed objects. This change breaks no existing tests and fixes the v2 regression. This change is also necessary for Kubernetes to upgrade from YAML v1 to v2 - see GoogleCloudPlatform/kubernetes#2490.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -31,9 +31,9 @@ type node struct {
 // Parser, produces a node tree out of a libyaml event stream.
 
 type parser struct {
-	parser  yaml_parser_t
-	event   yaml_event_t
-	doc     *node
+	parser yaml_parser_t
+	event  yaml_event_t
+	doc    *node
 }
 
 func newParser(b []byte) *parser {
@@ -193,10 +193,10 @@ type decoder struct {
 }
 
 var (
-	mapItemType = reflect.TypeOf(MapItem{})
-	durationType = reflect.TypeOf(time.Duration(0))
+	mapItemType    = reflect.TypeOf(MapItem{})
+	durationType   = reflect.TypeOf(time.Duration(0))
 	defaultMapType = reflect.TypeOf(map[interface{}]interface{}{})
-	ifaceType = defaultMapType.Elem()
+	ifaceType      = defaultMapType.Elem()
 )
 
 func newDecoder() *decoder {
@@ -488,8 +488,6 @@ func (d *decoder) sequence(n *node, out reflect.Value) (good bool) {
 	}
 	return true
 }
-
-
 
 func (d *decoder) mapping(n *node, out reflect.Value) (good bool) {
 	switch out.Kind() {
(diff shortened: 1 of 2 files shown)


niemeyer:
I don't understand why "" would result in calling an unmarshaler while an implicit "" would not. This is very inconsistent. I'm happy to fix the bug, but not like that.

Given the discussion in that ticket referenced above, I'm assuming are not interested in contributing to the project anymore, and will close this and fix the bug.

ghodss:
> I don't understand why "" would result in calling an unmarshaler while an implicit "" would not. This is very inconsistent. I'm happy to fix the bug, but not like that.

Because an empty string is not the same as null in Go. How do you propose fixing it instead?

> Given the discussion in that ticket referenced above, I'm assuming are not interested in contributing to the project anymore, and will close this and fix the bug.

As I commented, we are still using go-yaml/yaml for parsing YAML, but not for marshaling/unmarshaling. If the behavior were made to be more strongly consistent with how Golang marshals and unmarshals, we would consider switching back.

niemeyer:
Yes, you're right. I forgot "a:" actually implies a is null in YAML.

I'm happy to accept this in. For that, can you please:

a) Sign the contributor agreement at:

b) Preserve the original formatting of the line (no unnecessary parenthesis, as usual in Go) while adding the intended change?

Thanks for these.

niemeyer:
Ping?

ghodss:
My bad, I got caught up with a few things - I will definitely take care of
this within the next day. Thanks!

On Tue, Jan 20, 2015 at 10:51 AM, Gustavo Niemeyer <notifications@github.com

> wrote:
> 
> Ping?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/go-yaml/yaml/pull/57#issuecomment-70709482.

## 

Sam

ghodss:
I rebased and had some issues - I will update the commits here shortly as soon as I get a moment to fix it up.

ghodss:
@niemeyer, can you reopen the issue? I updated my `ghodss:fix-unmarshal-yaml` branch with the changes but I think the changes may not show up in this PR until it's reopened.

ghodss:
Also I believe I signed the agreement, let me know if it came through okay.

PR #59, opened 2014-12-02
Handle anonymous fields in structs and struct pointer inlining

jawher:
You can check the new tests to se the new handled cases

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -601,12 +601,7 @@ func (d *decoder) mappingStruct(n *node, out reflect.Value) (good bool) {
 			continue
 		}
 		if info, ok := sinfo.FieldsMap[name.String()]; ok {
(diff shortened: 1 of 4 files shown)


niemeyer:
Thanks for the submission.

I don't think we're ready to accept the inline change at this time. This would break compatibility with existent code, and introduce a redundant way to define inlining, which is currently done via the "inline" tag flag (for historical context, that implementation was done before Go's json pacakge supported inlining).

The struct pointer might be acceptable, but not as part of this change, and I'm also slightly unsure about whether it is properly implemented. When we have a pointer, we need to worry about initialization of nil pointers appropriately. I haven't checked in detail the correctness of the suggestion as we cannot accept this for the above reason.

PR #64, opened 2014-12-28
Don't fail if an Unmarshaler generates an intermediate error

nelhage:
When handling an `Unmarshaler`, the code slices any errors generated by the intermediate unmarshal call out of the parent `terrors` list. However, this results in the slice being non-nil, and so the root `Unmarshal` call will still fail with an empty error list, since it checks for non-nil, instead of nonzero length.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -2,13 +2,15 @@ package yaml_test
 
 import (
 	"errors"
-	. "gopkg.in/check.v1"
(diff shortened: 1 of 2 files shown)


niemeyer:
Sorry about that. This was already handled by #52. I was just too slow to respond due to the holidays season, but I should have taken care of it earlier.

PR #65, opened 2014-12-29
Bug fixed: inner function inside UnmarshalYAML can be called several times

coxx:
When implementing interface Unmarshaler sometimes it's useful to call inner function several times. In this case slice terrors initialized and despite the fact that there are no errors, terrors is not nil (but has zero length). Because it's safe to call len() on variable of slice type which can be nil, It's better to check `len(v) != 0`  than `v != nil`.

Here is working example
https://gist.github.com/coxx/ae6bc531573d72fb4d62

### DIFF
--- a/yaml.go
+++ b/yaml.go
@@ -90,7 +90,7 @@ func Unmarshal(in []byte, out interface{}) (err error) {
 		}
 		d.unmarshal(node, v)
 	}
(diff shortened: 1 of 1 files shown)


coxx:
Oops, there is pull request (https://github.com/go-yaml/yaml/pull/64) that doing exactly the same.

PR #71, opened 2015-02-04
README.md: field rename sample

gogolok:


niemeyer:
Can you please make the key just "e" in the yaml, and use "AltE" as a field name in the struct? This will be more clear.

gogolok:
@niemeyer I've changed the example

niemeyer:
Thanks, and sorry for not returning much earlier to this. I've added a renamed field to address #122, without introducing new data (this is in https://github.com/go-yaml/yaml/commit/2bf60357b89cbc6044dde700cf63bab94a615bf7)

Issue #63, opened 2014-12-22
Composite types do not unmarshal correctly (i.e. behavior differs from json.Unmarshal)


bcronin:
Looks like `yaml.Unmarshal()` and `json.Unmarshal()` handle composite types differently. The yaml unmarshalling does not appear to read the tags on the inner type.  Note `ValueA` is unmarshalled correctly in the json case, but not in the yaml case.

``` go
package main

import (
    "encoding/json"
    "fmt"

    "github.com/go-yaml/yaml"
)

type A struct {
    ValueA int `json:"a" yaml:"a"`
}
type B struct {
    A
    ValueB int `json:"b" yaml:"b"`
}

const jsonData = `{ "a" : 1, "b" : 2 }`
const yamlData = `
a : 1
b : 2
`

func main() {
    jdata := B{}
    ydata := B{}

    json.Unmarshal([]byte(jsonData), &jdata)
    yaml.Unmarshal([]byte(yamlData), &ydata)

    // $ go run main.go
    // json: {{1} 2}
    // yaml: {{0} 2}
    fmt.Printf("json: %v\n", jdata)
    fmt.Printf("yaml: %v\n", ydata)
}
```

advance512:
This works:

``` go
package main

import (
    "encoding/json"
    "fmt"

    "gopkg.in/yaml.v2"
)

type A struct {
    ValueA int `json:"a" yaml:"a"`
}
type B struct {
    A      `yaml:",inline"`
    ValueB int `json:"b" yaml:"b"`
}

const jsonData = `{ "a" : 1, "b" : 2 }`
const yamlData = `
a : 1
b : 2
`

func main() {
    jdata := B{}
    ydata := B{}

    json.Unmarshal([]byte(jsonData), &jdata)
    yaml.Unmarshal([]byte(yamlData), &ydata)

    // $ go run main.go
    // json: {{1} 2}
    // yaml: {{1} 2}
    fmt.Printf("json: %v\n", jdata)
    fmt.Printf("yaml: %v\n", ydata)
}
```

advance512:
(I just added a tag on the A anonymous field.) Whether or not this is "good behavior" or not depends. I agree that, generally, go-yaml should work in a similar way to go-json. In this case, however, I don't really like the way go-json works - it just seems too "magicky".

That being said, adding `yaml:",inline"` isn't a pretty solution either.

advance512:
[...]

PR #77, opened 2015-02-11
Inline anonymous structs.

robx:
`encoding/json` inlines anonymous structs, this brings yaml in line.

There's one renaming commit in here, too, would be happy to split that out if you prefer.


robx:
Compare http://play.golang.org/p/a65jCwme9y

niemeyer:
Please see the discussion here: https://github.com/go-yaml/yaml/issues/63

PR #78, opened 2015-02-11
Clearer names.

robx:
`fields` is quite misleading next to `field`.


niemeyer:
Sorry, I appreciate the intent, but when it's just about a very subtle difference about variable naming, let's please not bikeshed over it.

robx:
Whatever.

niemeyer:
Right, that's exactly the spirit. Renaming variables for questionable reasons is not worth your time or mine.

Issue #13, opened 2014-06-07
Feature request: json.RawMessage equivalent


michaelsauter:
It would be great to have sth. similar to http://golang.org/pkg/encoding/json/#RawMessage.

I'm using go-yaml in http://github.com/michaelsauter/crane and would like to unmarshal the JSON/YAML partially (to provide defaults for objects inside arrays).

GeorgeMac:
I would like to +1 this. Would be very useful for writing config files in my projects!

fgrosse:
Yeah I am missing that too so I can delay the actual decoding of some fields.
Is there a known work around? I don't get how the Unmarshaller interface[1](http://godoc.org/gopkg.in/yaml.v2#Unmarshaler) is supposed to work.

ghost:
+1 for yaml.RawMessage

<b>edit</b>

The `yaml.MapSlice` structure seems to be a much better solution than `json.RawMessage`

missedone:
+1 for yaml.RawMessage

niemeyer:
The problem with yaml.RawMessage is that unlike JSON, YAML is context-dependent. One cannot properly encode a piece of raw YAML without knowing what the surrounding data is, due to indentation.

Can the +1 voters please detail a bit how this is intended to be used that isn't solved by the current system?  Maybe we can design something else that solves the problem.

ghost:
@niemeyer The problem I had intended to solve using `RawMessage` I had later solved using `MapSlice`. I am relatively new to golang, so this wasn't immediately obvious to me.

I think most problems people intend to solve with a `json.RawMessage` equivalent may be solved with `yaml.MapSlice`, but there isn't much documentation or example usages of it. If there was perhaps more of  this, people will find that they do not need a `yaml.RawMessage`.

missedone:
@niemeyer In my case, I have a lot of yaml configuration files, each includes some common definition and plugin settings, for plugin settings it's variant from plugin to plugin, i want to delay the parse and pass the RawMessage to leave the plugin implementation do it.
i know i can extract the plugin settings into separate yaml file, but it's really annoying i need to reference them in the parent yaml file.

fgrosse:
I have the exact same use case as @missedone

niemeyer:
Yeah, it should be easy to use a MapSlice, and one can even marshal it back into yaml and then unmarshal it again to get the exact semantics of a RawMessage. That said, I'll keep this issue open so we can consider the idea of a RawMessage-like feature further.

[...]

PR #86, opened 2015-02-25
decode: add mapping key transform hook

crawford:
This hook allows the user to apply a custom transform to mapping keys
during unmarshalling.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -32,13 +32,15 @@ type node struct {
 // Parser, produces a node tree out of a libyaml event stream.
 
 type parser struct {
-	parser yaml_parser_t
-	event  yaml_event_t
-	doc    *node
+	parser    yaml_parser_t
+	event     yaml_event_t
+	doc       *node
+	transform transformString
 }
 
-func newParser(b []byte) *parser {
-	p := parser{}
+func newParser(b []byte, t transformString) *parser {
+	p := parser{transform: t}
+
 	if !yaml_parser_initialize(&p.parser) {
 		panic("failed to initialize YAML emitter")
 	}
@@ -177,7 +179,10 @@ func (p *parser) mapping() *node {
 	p.anchor(n, p.event.anchor)
 	p.skip()
 	for p.event.typ != yaml_MAPPING_END_EVENT {
-		n.children = append(n.children, p.parse(), p.parse())
+		key := p.parse()
+		key.value = p.transform(key.value)
+		value := p.parse()
+		n.children = append(n.children, key, value)
 	}
 	p.skip()
 	return n
--- a/decode_test.go
+++ b/decode_test.go
@@ -812,6 +812,23 @@ func (s *S) TestUnmarshalerRetry(c *C) {
 	c.Assert(su, DeepEquals, sliceUnmarshaler([]int{1}))
 }
 
+func (s *S) TestUnmarshalWithTransform(c *C) {
+	data := `{a_b: 1, c-d: 2, e-f_g: 3, h_i-j: 4}`
+	expect := map[string]int{
+		"a_b":   1,
+		"c_d":   2,
+		"e_f_g": 3,
+		"h_i_j": 4,
+	}
+	m := map[string]int{}
+	yaml.UnmarshalMappingKeyTransform = func(i string) string {
+		return strings.Replace(i, "-", "_", -1)
+	}
+	err := yaml.Unmarshal([]byte(data), m)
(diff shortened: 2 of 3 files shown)


crawford:
Bump. Is this not the appropriate place to submit patches?

niemeyer:
Yes, this is the appropriate place to submit patches, and my apologies for the lack of feedback. It wasn't immediately obvious what to do about this, and then it just got out of my critical path.

This doesn't sound like functionality that should be inside the yaml package. If you want to do some specific marshaling to fix some situation you have, use a custom marshaler for that specific case via the [yaml.Unmarshaler](http://gopkg.in/yaml.v2#Unmarshaler) interface. If you want arbitrary manipulation of keys, then this can be done at a map level. In other words, unmarshal the document into a map and then do whatever is necessary to it.

crawford:
The particular case that I had in mind was unmarshalling cloud-configs, where `-` and `_` can be used interchangeably. Using yaml.Unmarshaler seems like it would be very cumbersome. Unless I am mistaken, all non-leaf elements would need to have a custom Unmarshaler, in which it would need to unmarshal into a map as an intermediate step. I want to avoid using a map altogether since that basically throws away the type system. I depend heavily on the structured types in order to validate user input.

I agree that applying transforms to YAML keys is fairly uncommon, but I couldn't think of a better way to solve this issue.

niemeyer:
Having loose keys as suggested seems indeed like an edge case, and it seems sensible to not add functionality until we feel like there are more common cases that would benefit from the overhead.

For the time being, I'd suggest doing a pre-process step: unmarshal into a map, manipulate as necessary, remarshal, and unmarshal into the final location. This is surely a suboptimal workaround, but it feels worth paying the price where the cumbersome strategy is being observed before we have a better reason to have everyone paying the price.

crawford:
My first implementation was almost exactly what you suggested. This doesn't work unfortunately because of YAML's type coercion. Consider the following example:

``` yaml
person:
  name: crawford
  nick: off
  phone: 0123456701
```

If I were to unmarshal this into a map I would get the following:

``` go
map[interface{}]interface{} {
    "person": map[interface{}]interface{} {
        "name":  "crawford",
        "nick":  "false",
        "phone": 21913025,
    },
}
```

instead of:

``` go
Person {
    Name:  "crawford",
    Nick:  "off",
    Phone: "0123456701",
}
```

Yes this is a little contrived, but it illustrates two problems we have actually run into. Without a type to unmarshal into, and because YAML doesn't require quotes around strings, the parser chooses the most likely type (a boolean for the nick and an integer for the phone number) which is incorrect in this particular case. We need to hint to the parser which type to use when unmarshalling each value. Of course, the easy solution is to just write correct YAML in the first place, but unfortunately this is user provided.

niemeyer:
I see, there's a problem indeed. To be clear, though, the issue here is that the implementation seems to be trusting on invalid YAML files. It wouldn't be the first time, but accounting it as "the most likely type" and saying that the YAML doesn't require quotes to interpret "off" as a string isn't correct. There's only one type the YAML should be read as, and the package is doing that job correctly.

Loading it in Python, for example:

```
>>> yaml.load("nick: off")
{'nick': False}
```

So if users are providing that file expecting it to be a string, that's a broken YAML file which will be handled incorrectly by other implementations.

Despite that, one thing we can do to fix this and which I'd be happier pursuing, is to introduce the notion of raw values, and make sure they marshal out unchanged. This is being discussed in #13.

crawford:
> saying that the YAML doesn't require quotes to interpret "off" as a string isn't correct

I never said that. I said "YAML doesn't require quotes around strings" which is absolutely true.

> There's only one type the YAML should be read as, and the package is doing that job correctly.

But there is a difference in behavior between unmarshalling into a typed container vs an untyped container. This is the underlying problem.

That being said, RawMessage is the correct solution. I'm fine tossing this in favor of #13.

niemeyer:
> I never said that. I said "YAML doesn't require quotes around strings" which is absolutely true.

You said that _"and because YAML doesn't require quotes around strings"_ while describing your case which was with the "off" string. That can't be "absolutely true" when it is false for the very case you're talking about.

> But there is a difference in behavior between unmarshalling into a typed container vs an untyped container. This is the underlying problem.

The real underlying problem is that invalid YAML is being used. This will be an issue with any correct YAML parser you use.

> That being said, RawMessage is the correct solution. I'm fine tossing this in favor of #13.

Sounds good. Thanks for the conversation.

PR #93, opened 2015-03-31
Automatically decode embedded structs "inline"

mvo5:
By adding "inline" automatically for embededed struct all (visible)
values from the embedded struct are read (this is what json is
doing). Fixes LP: #1191814


niemeyer:
This would break existing applications that rely on the current behavior. We can eventually switch off to do that by default, but not without a major version change.

mvo5:
Thanks, I see the risk here. I created a small branch that adds a example how to do the decoding instead to help users to find it more easily.

PR #95, opened 2015-04-13
Use an svg icon for the documentation.

maruel:
Concatenate API documentation and API stability into one section.
Switch style for package name in Installation section.

Compare https://github.com/go-yaml/yaml/tree/v2 and https://github.com/maruel/yaml/tree/v2 for the look. I like the icon, it's easy to recognize and clicking on it always leads to the expected result: documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,21 +20,16 @@ supported since they're a poor design and are gone in YAML 1.2.
 Installation and usage
 ----------------------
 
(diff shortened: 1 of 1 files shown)


niemeyer:
The fact the import path may be opened in a browser in a useful way is lost in this change.

PR #96, opened 2015-04-14
Add case for float zero-values

gyuho:


gyuho:
Hello, thanks for the great YAML package.

I just ran into an issue of **NOT** omitting the zero float values.
It happens when I `Unmarshal` the data with zero float64 values and `Marshal` back to YAML format.

The master branch does **NOT** check the zero values of float types. (https://github.com/go-yaml/yaml/blob/v2/yaml.go#L328)

Could we add a line to check zero float values? Otherwise, 0 or 0.0 on float types will never be omitted when `Marshal`ing.

Please let me know.

Thanks,

niemeyer:
Hi, sorry for not getting back to you on this timely. I'll fix the issue.

gyuho:
Thanks @niemeyer !

niemeyer:
Tested and fixed in 7ad95dd. Thanks for the report.

PR #102, opened 2015-05-19
Fix bug in `yaml_emitter_analyze_scalar()` when the string starting with multi-byte character

Songmu:
`width(value[0])` always returns the width of first character of the string. It should be `width(value[i])`.

### DIFF
--- a/emitterc.go
+++ b/emitterc.go
@@ -1019,7 +1019,7 @@ func yaml_emitter_analyze_scalar(emitter *yaml_emitter_t, value []byte) bool {
 
 	preceeded_by_whitespace = true
 	for i, w := 0, 0; i < len(value); i += w {
-		w = width(value[0])
+		w = width(value[i])
 		followed_by_whitespace = i+w >= len(value) || is_blank(value, i+w)
 
 		if i == 0 {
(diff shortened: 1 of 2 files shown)


daiksy:
:+1:

yuuki:
:+1:

Sixeight:
:+1:

Songmu:
@niemeyer How about it?

This is trivial, careless and frequent index bug, but it is
serious bug in multi-byte environment.

For example, `map[string]string{"a": "你好 #not-a-comment"}` is marshaled to `a: 你好 #not-a-comment`,
but, this result `a: 你好 #not-a-comment` is unmarshaled to `map[string]string{"a": "你好"}` and `#not-a-comment` is removed.

niemeyer:
Thanks for all the feedback. This is definitely going in.

niemeyer:
@Songmu If you haven't yet done so, can you please sign the contributors agreement so I can integrate this:

http://www.ubuntu.com/legal/contributors

It's a non-draconian license agreement which should be a straightforward read, and can be easily signed online.

Songmu:
It seems to be done. Could you check it?

niemeyer:
Thanks!

Songmu:
Thank you!

PR #103, opened 2015-05-25
Fix for non-specific tags parsing

a-chernykh:
Make it acting like libyaml which includes length of the head when testing for emptiness: https://github.com/yaml/libyaml/blob/master/src/scanner.c#L2582

### DIFF
--- a/scannerc.go
+++ b/scannerc.go
@@ -611,7 +611,7 @@ func yaml_parser_set_scanner_tag_error(parser *yaml_parser_t, directive bool, co
 	if directive {
 		context = "while parsing a %TAG directive"
 	}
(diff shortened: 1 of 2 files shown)


emosbaugh:
+1

scannerc.go
```diff
@@ -1996,14 +1997,15 @@ func yaml_parser_scan_tag_uri(parser *yaml_parser_t, directive bool, head []byte
 			}
 		} else {
 			s = read(parser, s)
+			length++
```
rogpeppe, line 2000: This doesn't work when the tag name consists entirely of URL-escaped characters.

rogpeppe (approved):
Thanks for this. LGTM except that it doesn't work when the tag name consists entirely of % escaped characters. I'll merge this and then fix that separately.

PR #112, opened 2015-07-06
Expose line and column number of parsed elements

### LINKED ISSUE #108: Exposing line numbers in the API
I'd like to be able to get line/col numbers for each YAML element. I'm contributing to a project using YAML as input to a compiler, and we want to show position information in compiler error messages.

I'm interested in adding that feature if you'd take pull requests for it.

It looks like go-yaml tracks all the positions internally, but doesn't expose them through the API. The hard part seems to be designing an API to expose positions. I have a few thoughts, but I figured I'd see if there is interest before going too far into the details.

clarete:
Here's my stab at exposing line and column numbers to the mapped structs.

The `linenum` and `colnum` flags were introduced to mark fields that
will save the line number and column number of an element.

The field should not really map to any field names on the YAML document,
since the parser won't really save it. Once line or column number is
saved to a field that carries the flags mentioned here, the parser is
done with that field.

PS.: I changed the import path for the `yaml` package on both `{encode,decode}_test.go`
files, because I needed to run the tests using my fork that has a different address. I guess
that makes everyone else's life easier too. Let me know if I should break that change into
a different patch.

Might fix #108

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -586,7 +586,7 @@ func (d *decoder) mappingSlice(n *node, out reflect.Value) (good bool) {
 			d.merge(n.children[i+1], out)
 			continue
 		}
-		item := MapItem{}
+		item := MapItem{Line: n.children[i].line, Col: n.children[i].column}
 		k := reflect.ValueOf(&item.Key).Elem()
 		if d.unmarshal(n.children[i], k) {
 			v := reflect.ValueOf(&item.Value).Elem()
--- a/decode.go
+++ b/decode.go
@@ -616,6 +616,17 @@ func (d *decoder) mappingStruct(n *node, out reflect.Value) (good bool) {
 		elemType = inlineMap.Type().Elem()
 	}
 
+	for _, f := range sinfo.FieldsMap {
+		if f.LineNum {
+			field := out.Field(f.Num)
+			field.SetInt(int64(n.line))
+		}
+		if f.ColNum {
+			field := out.Field(f.Num)
+			field.SetInt(int64(n.column))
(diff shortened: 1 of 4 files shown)


niemeyer:
This change breaks current code using MapItem and thus MapSlice, so we have it in as-is. It's also concerning that this data won't be used when marshaling the items back.

Can you also please describe what is the use case for this?

yobert:
I think this is a useful feature. For me I would use it for reporting an error in a YAML config file. Not a YAML syntax error, but an error specific to my config syntax.

What if we did a modified version of this where it left MapItem and MapSlice alone, but still allowed the custom struct tag mode? Then it's a bit more explicit and it doesn't break any compatibility.

niemeyer:
Sure, I'm certainly not against the idea of exposing the feature somehow, if we find a good way.

kyessenov:
This would really help with usability when dealing with large structured YAML files, such as Kubernetes manifests. Line numbers, FTW!

niemeyer:
v3 will expose the ability to decode into an intermediate state with line/column numbers, and encode with the main style (order, flow, etc) preserved, including comments.

Sneak peek: https://twitter.com/gniemeyer/status/1073638853225996288

tisba:
That sounds quite interesting actually. Is there an issue or PR we can follow, @niemeyer?

tisba:
Note to self (and whoever might be interested): v3 was released, https://blog.ubuntu.com/2019/04/05/api-v3-of-the-yaml-package-for-go-is-available

PR #117, opened 2015-08-11
Add support for alternative timestamp formats

anthonybishopric:
yaml.org describes several alternative formats for timestamps not
covered by the default implementation of `time.UnmarshalText()`. These
are handled explicitly in this change in resolve.go

Timestamps will only be collected if there is an explicit timestamp
tag (ie, not `!!str`) or if implicit type detection is enabled.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -453,6 +453,11 @@ func (d *decoder) scalar(n *node, out reflect.Value) (good bool) {
 			out.SetFloat(resolved)
 			good = true
 		}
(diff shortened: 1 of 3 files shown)


rogpeppe:
Thanks very much for doing this. Unfortunately it doesn't go as far as the
spec (http://yaml.org/type/timestamp.html) allows, and it breaks
the tests, but it's still useful - I've used it as the basis of new PR.
See PR #273.

PR #122, opened 2015-08-29
add example for using tag

hvnsweeting:


niemeyer:
Thanks for the change. Indeed it'd be nice to have a better example of tag renaming in the documentation, but we co do that with a more trivial change. I've done that in 2bf6035. Please let me know if it solves your needs.

hvnsweeting:
yes, it's good. Thanks

PR #124, opened 2015-09-15
preserve crlf when parsing literal blocks

coryb:
While scanning the yaml stream go-yaml will drop the raw line termination and always replace it with '\n'.  In general this is fine unless you have a literal block with crlf that you need preserved:

``` yaml
stuff: |
  this is^M
  a line^M
```

Here is a sample program that shows the problem:

``` yaml
package main

import (
        "fmt"
        "gopkg.in/yaml.v2"
)

func main() {
        data := make(map[string]string)
        yaml.Unmarshal([]byte("multiline: |\n  a\r\n  b\r\n  c\r\n"), &data)
        fmt.Printf("multiline: %#v\n", data["multiline"])
}
```

``` bash
$ go run yaml-test.go
multiline: "a\nb\nc\n"
```

You can see the \r\n are replaced with \n

if we swap the `gopkg.in/yaml.v2` with my patched version `gopkg.in/coryb/yaml.v2` and run it again we will get correct output:

``` bash
$ go run yaml-test.go
multiline: "a\r\nb\r\nc\r\n"
```

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -193,6 +193,12 @@ var unmarshalTests = []struct {
 		map[string]string{"scalar": "\nliteral\n\n\ttext\n"},
 	},
 
+	// Literal block scalar with crlf
+	{
+		"scalar: | # Comment\n\r\n literal\r\n\r\n \ttext\r\n\n",
+		map[string]string{"scalar": "\r\nliteral\r\n\r\n\ttext\r\n"},
+	},
+
 	// Folded block scalar
 	{
 		"scalar: > # Comment\n\n folded\n line\n \n next\n line\n  * one\n  * two\n\n last\n line\n\n",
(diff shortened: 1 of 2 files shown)


niemeyer:
The current behaviorismo is part of the specification. Please ser section 5.4 on normalization of line-break characters.

niemeyer:
And that's what I get for typing on a phone. Sorry for the typos.

coryb:
Thanks for the feedback, and spec section.  It is surprising as it makes YAML useless as a data exchange format since it does not preserve the original data.  I guess I will maintain my fork.

niemeyer:
The spec has several defects, most notably bloat, but I would say this is one thing it got right. It is definitely a data exchange format, but luckily not one that preserves invisible EOL characters as something meaningful. If you want bit by bit preservation, encode the data, quote it, or simply pick a binary data exchange format.

PR #126, opened 2015-09-25
Add linenum and colnum struct field support

yobert:
Another attempt at exposing line/column information during parsing. Similar to pull request #112 by @clarete but a little simpler. (Though it may not solve his use case).

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -559,6 +559,17 @@ var unmarshalTests = []struct {
 		"a: []",
 		&struct{ A []int }{[]int{}},
 	},
+
+	// Line number information from parser
+	{
+		"\n\n\n\n\n\n\n\n\n\na: one\nb: two\nc: three\n",
+		&struct {
+			A    string
+			B    string
+			Line int ",linenum"
+			Col  int ",colnum"
+		}{"one", "two", 11, 1},
+	},
 }
 
 type M map[interface{}]interface{}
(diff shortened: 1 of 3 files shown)


decode_test.go
```diff
@@ -559,6 +559,17 @@ var unmarshalTests = []struct {
...
+
+	// Line number information from parser
+	{
+		"\n\n\n\n\n\n\n\n\n\na: one\nb: two\nc: three\n",
+		&struct {
+			A    string
+			B    string
+			Line int ",linenum"
+			Col  int ",colnum"
+		}{"one", "two", 11, 1},
```
niemeyer, line 571: It's not clear what the line number and column number information mean here. There is a struct with multiple fields, and only a single exposed number information which maps to the first field apparently. What's are the semantics, and how is it useful?

yobert, line 571: It would be the line and column of where in the YAML that object "started". It's not perfectly granular, and only works for objects, so I understand completely if this is too hacky for you to want to add to go-yaml.  My use case is I have a file that has an array of several thousand objects. So when I find an error, I can report on what line that object began.

niemeyer:
Yeah, sorry, this feels like an unexpected meaning for line/column which people will be confused about. I'm closing this PR for the time being. Feel free to discuss further.

yobert:
Yeah, makes sense. This fix is too specific to our use cases. I wish there was a clean way to get the line/column down to the level of a string inside an array of strings...

yobert:
If anyone else stumbles on this thread, for now I'm going to maintain github.com/yobert/yaml for this behavior.

PR #129, opened 2015-10-06
Support emitting comments

DirectXMan12:
This commit introduces support for YAML comments.  Currently,
comments are placed on the line before mapping keys (or potentially
sequence items).  Comments placed at the end of lines are not supported.

Comments may be created either by setting `yamlComment:"some comment"`
in a struct field tag, or by setting the `Comment` field of `CommentedMapItem`
(and using `CommentedMapSlice` objects instead of `MapSlice` objects).
Empty comments are skipped.  Currently, adding comments above list items
is not supported (but you can still add comments to struct fields in list items).

I introduced `CommentedMapItem` and `CommentedMapSlice` in order to not break the API
by changing the signature of `MapItem`.  If you'd rather bump the API version to have a bit cleaner
code under the hood, I can change it so that `MapItem` just has a `Comments` field instead.

### DIFF
--- a/apic.go
+++ b/apic.go
@@ -327,6 +327,15 @@ func yaml_scalar_event_initialize(event *yaml_event_t, anchor, tag, value []byte
 	return true
 }
 
+// Create COMMENT.
+func yaml_comment_event_initialize(event *yaml_event_t, value []byte) bool {
+	*event = yaml_event_t{
+		typ:   yaml_COMMENT_EVENT,
+		value: value,
+	}
+	return true
+}
+
 // Create SEQUENCE-START.
 func yaml_sequence_start_event_initialize(event *yaml_event_t, anchor, tag []byte, implicit bool, style yaml_sequence_style_t) bool {
 	*event = yaml_event_t{
--- a/emitterc.go
+++ b/emitterc.go
@@ -209,6 +209,29 @@ func yaml_emitter_increase_indent(emitter *yaml_emitter_t, flow, indentless bool
 
 // State dispatcher.
 func yaml_emitter_state_machine(emitter *yaml_emitter_t, event *yaml_event_t) bool {
+	// comments do not change the state, so deal with them separately
+	if event.typ == yaml_COMMENT_EVENT {
+		switch emitter.state {
+		// on the first item of a mapping, we need to indent
+		case yaml_EMIT_BLOCK_MAPPING_FIRST_KEY_STATE:
+			return yaml_emitter_emit_comment(emitter, event, true)
+		// place comments on their own line, and never inside
+		// a flow-style mapping or sequence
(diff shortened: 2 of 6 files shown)


niemeyer:
Hi there,

I'm missing some high-level context for what this addressing and how it is addressing it.

Can you point me to the section of the specification that describes what you are doing, and how you are implementing support for it, including some more clear examples?

Thanks!

DirectXMan12:
Intro to comments: http://yaml.org/spec/1.2/spec.html#id2767100
Comment specification: http://yaml.org/spec/1.2/spec.html#id2780069 (this commit, for the moment, only supports emitting comments on their own line, not at the end of a line)
See also http://yaml.org/spec/1.2/spec.html#id2762313

Comments are useful in that they allow extra "human-readable" information to be attached without changing the structure of the document.  For instance, suppose a program returns a piece of stored configuration.  Using comments, it could add detailed descriptions of each of the fields in the configuration, in case the person reading them was not fully familiar with the format.  Additionally, comments could be used when "interactively" editing configuration -- the configuration is opened in an editor, the user saves/quits, the configuration is validated, and the editor is reopened with comments above any lines with issues.

The approach is as such:
- At the low-level (in `emitterc.go`), I introduced a new event type -- `yaml_COMMENT_EVENT`.  This event does not change the current state, and only generates any output when waiting for a key (or list item) and in the block style (you can't have a comment in the middle of a flow-style object, and also can't have a comment in between a key and a value).
- At the high levels, comments may be inserted either through the use of a struct field tag (`yamlComment`) for when static comments are sufficient, or through the use of an extend variant of the `MapItem`: `CommentedMapItem`.  This support "dynamic" comments, such as in the validation example.  When a user attempts to marshal a `CommentedMapItem`, the encoder checks it for a non-empty comment.  If it finds a non-empty comment, it emits a comment event.  In either case, it then proceeds as usual for a `MapItem` (I didn't modify the existing MapItem so as to preserve API backwards-compatibility).

Technically, the low-level code supports adding comments above list items.  However, I didn't see an easy way to expose this to the high level.

It shouldn't be too hard to add support for postfix comments -- I can add this if you want.  However, there's not really a good way to support them from the high-level without introducing more configuration and/or tags (and then things start to get annoyingly verbose), and this functionality is not strictly necessary for what I'm trying to accomplish (which is, basically, the two example cases described above).

niemeyer:
We can certainly support comment manipulation, but we'll need to design this a little bit better first.

Namingly, we need to ignore the internal implementation completely and look from the perspective of someone using the feature both to consume comments and to produce them back into the expected locations. We also need to outline which cases we'll try to support when writing, and which cases are going to cause clear havoc in the output.

I'm closing this PR but I'd be happy to see an issue opened to discuss these details.

PR #131, opened 2015-10-07
Added Encode() and Decode() to make duck-typing with json- and xml-counterparts easier 

oec:
Hi,

In my current project I needed to be able to handle Encoders and Decoders for any of the formats xml, json and yaml more easily and ended up declaring a corresponding interface type and implementing Encode() and Decode() in the yaml-package.  That way, I can pass En-/Decoders around.  See http://play.golang.org/p/JGD2dhk7ZU for a trivial example.

Maybe you'll find it usefull.

Cheers,
  Özgür

### DIFF
--- a/yaml.go
+++ b/yaml.go
@@ -9,6 +9,8 @@ package yaml
 import (
 	"errors"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"reflect"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


jeffjen:
Its been awhile since the owner had any comment on this.

The document states that Unmarshal decodes the **first** document found.  I wanted the Decoder and Encoder interface to match semantics of **encoding/json** package's.

I do have a change that will address this problem, but it will take more time for me to figure out how to not deplete user's memory during decode phase in a real **streaming** scenario, where the input buffer is expected to be unbounded.

niemeyer:
The functionality is sane, but the names should be Decoder and Encoder, and we need proper tests.

If we don't get @oec to fix that soonish, I'll add the equivalent functionality myself over the next few days.

oec:
@niemeyer, I can change the names and write some tests until Wednesday.  @jeffjen, handling multiple documents would break semantics with encoding/xml, AFAIK.   Maybe a switch in the (yaml-)en/decoder would be appropriate to give control over the behavior to the consumer of the API.

niemeyer:
One of the most important points of having that interface is fixing streaming of multiple documents, as @jeffjen mentions. The other would be having custom flags that can act per Decoder/Encoder instance, so we can have behavior changes (e.g. non-lowercase field names) without breaking existing code.

niemeyer:
@oec About breaking semantics of encoding/xml, I'm not quite sure about what you're describing, but have a look at the [example of encoding/json](https://golang.org/pkg/encoding/json/#Decoder). We want the same for yaml, except we'd respect the spec in terms of how to split docs.

oec:
@niemeyer, @jeffjen, I got your point and it should easily be possible to implement that behavior.

The semantic issue I ran into was different: decoder.xml requires the top-level tag to be part of the stream (in your example it would be < Message >), see http://play.golang.org/p/s3MCpzvsBG.  But this is independent of what you want.

oec:
@niemeyer, how shall we proceed?  I think I won't be able to implement streaming support and tests before Friday.

niemeyer:
Let's just close this PR and I'll try to get to it over the next few days then.

Thanks for your help here @oec.

jeffjen:
@oec I missed the discussion!!  Anyway I submitted a pull request #163  to add and Deocder/Encoder struct to handle streaming cases, but I didn't wrote test for it.

Issue #75, opened 2015-02-10
Fails to parse non-specific node tags


svend:
``` go
package main

import (
    "fmt"
    "log"

    "gopkg.in/yaml.v2"
)

var data = `
- "12"
- 12
- ! 12
`

func main() {
    m := make(map[interface{}]interface{})

    err := yaml.Unmarshal([]byte(data), &m)
    if err != nil {
        log.Fatalf("error: %v", err)
    }
    fmt.Printf("--- m:\n%v\n\n", m)
}
```

Results in `2015/02/10 12:47:50 error: yaml: line 3: did not find URI escaped octet`.

According to the [YAML 1.2 non-specific tag documentation](http://yaml.org/spec/1.2/spec.html#id2784064), this YAML should parse to:

``` yaml
%YAML 1.2

---
!!seq [
  !<tag:yaml.org,2002:str> "12",
  !<tag:yaml.org,2002:int> "12",
  !<tag:yaml.org,2002:str> "12",
]
```

a-chernykh:
This bugs me too :+1:

hashmap:
Pull request has been submitted https://github.com/go-yaml/yaml/pull/133

emosbaugh:
+1

geofffranks:
It's been a year, and this hasn't been merged, or really discussed. Is there a timeline for getting this fixed?

nmiyake:
+1 -- ran into this issue as well and confirmed that the PR that is currently open resolves this issue. This seems strictly better/more correct from a correctness standpoint. Is there any reason that this issue/PR isn't being advanced?

niemeyer:
Yeah, sorry about that. As noticed elsewhere, this last year of mine was specially busy with work elsewhere. But we've just made the critical release a couple of weeks ago and I'm slowly picking up on the work I left behind. go-yaml is next in line.

igorwwwwwwwwwwwwwwwwwwww:
Just ran into this too, any updates?

rogpeppe:
Fixed by #103

PR #133, opened 2015-10-13
Fix parsing of non-specific node tags (issue #75)

hashmap:
Non-specific tag disables tag resolution, forcing the node to be
interpreted as `tag:yaml.org,2002:seq`, `tag:yaml.org,2002:map`, or
`tag:yaml.org,2002:str`

This fix pevents `yaml_parser_scan_tag_uri` from raising an error if
`suffix` is empty. This case is handled by function `yam_parser_scan_tag`:
https://github.com/go-yaml/yaml/blob/v2/scannerc.go#L1882
so this behaviour is expected.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -403,6 +403,12 @@ var unmarshalTests = []struct {
 	}, {
 		"%TAG !y! tag:yaml.org,2002:\n---\nv: !y!int '1'",
 		map[string]interface{}{"v": 1},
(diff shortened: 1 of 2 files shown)


emosbaugh:
+1

HeavyWombat:
+1

niemeyer:
This has been fixed a long while ago.

PR #138, opened 2015-10-21
Changed map[interface{}]interface{}{} to map[string]interface{}

### LINKED ISSUE #137: Change default map to map[string]interface{} instead map[interface{}]interface{}
https://github.com/go-yaml/yaml/blob/v2/decode.go#L199

After reading from a yml file I can't marshal as json since I get `json: unsupported type: map[interface {}]interface {}`. Is it possible to change the default map to map[string]interface{} instead map[interface{}]interface{}?

aanm:
Fixes #137 

Signed-off-by: André Martins aanm90@gmail.com


niemeyer:
See referenced issue.

Issue #136, opened 2015-10-18
decoder silently ignores fields in the yaml-input for which no field in the struct-type exists. 


oec:
If I have a struct

```
   type T struct {
      A string
   }
```

and use the following string as input:

```
   data := "a: foo\nb: bar\n"
```

then the call to yaml.Unmarshal

```
  t := &T{}
  e := yaml.Unmarshal(data, t)
  ...
```

will succeed without error and thereby silently ignore field "b" in the input.

I consider this behaviour a bug, as it doesn't allow f.e. to catch misspellings in config-files.

In my project I fixed this with the following patch for decode.go:

```
--- decode.go   2015-10-18 17:55:27.332842372 +0200
+++ my/decode.go    2015-10-18 17:14:26.421462484 +0200
@@ -640,6 +640,9 @@
            value := reflect.New(elemType).Elem()
            d.unmarshal(n.children[i+1], value)
            inlineMap.SetMapIndex(name, value)
+       } else {
+           d.terrors = append(d.terrors, fmt.Sprintf("line %d: no such field '%s' in struct '%s'", n.line+1, name, out.Type()))
+           return false
        }
    }
    return true

```

oec:
Of course, another - more complex - patch would be to implement something similar as the ",any"-tag in encoding/xml.  Unmarshaling into struct _without_ a field with that tag would silently ignore extra fields in the input (as it is now).  A field of type map[string]interface{} with the tag ",any" could serve as a catch-all and consumers could decide what to do with it.

abligh:
+1 - at least an option to error this would be good.

wfarner:
Are there any plans to accept the above patch, possibly with added plumbing to make the behavior optional/non-default?  Happy to pitch in.

niemeyer:
Erroring in that case would be unacceptable, but we can do something similar to what I've used elsewhere: support the inline tag in maps too. This is useful in a number of cases as it allows mixing previously known key with custom ones, and if you want to error when the map is not empty that's easy too.

wfarner:
@niemeyer can you elaborate?  I'm not following how inlining maps helps to address the use case @oec demonstrated in the original issue filing.

vincentbernat:
Is there a known workaround for this bug/feature?

douglaswth:
@vincentbernat there's my branch with these changes: #146.

vincentbernat:
[...]

PR #146, opened 2015-11-12
Don't ignore unspecified fields (from #136)

douglaswth:
This is based on @oec's comment in #136.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -640,6 +640,9 @@ func (d *decoder) mappingStruct(n *node, out reflect.Value) (good bool) {
 			value := reflect.New(elemType).Elem()
 			d.unmarshal(n.children[i+1], value)
 			inlineMap.SetMapIndex(name, value)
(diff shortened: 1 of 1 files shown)


dhiltonp:
+1

jdonboch:
+1

jdonboch (approved):

niemeyer:
We can't change the default behavior like this, but we can have a mode on a `Decoder` that enables it.

gemcdaniel:
I would like to see a mode on a `Decode` for strict decoding that would be thread-safe.

jdonboch:
@niemeyer We opened another PR #210 that doesn't change the default behavior and adds the concept of a moded Decoder

niemeyer:
Sounds good, thanks for that.

My most pressing deadlines elsewhere are due this month, so you can expect some good amount of love in go-yaml soon.

Issue #148, opened 2015-11-25
Default decoding/encoding behaviour is surprising for a Go library


samsalisbury:
I was very surprised to find that `Unmarshal` _by default_ assumes that my exported, Uppercased struct field names should only be unmarshalled if they are different in the YAML file (that is, if they are lowercased there).

Surely, if a user wants to use a different field name in their YAML vs Go code, they should use the field tags to indicate this. Having this magic case switching _just happen_ feels very odd, especially when this is compared with the built-in `encoding/json` package, which does not do any such magic.

I guess it's too late to change this for v2, as it would be a breaking change. However adding a func to disable this behaviour on an opt-in basis would be very useful indeed.

Note that this is a different issue from https://github.com/go-yaml/yaml/issues/123 because once you have used the field tags to specify specific names, I don't see why the parser should perform any additional case-insensitive magic unless somehow told to.

niemeyer:
Please see #120 for motivation details. Will comment on the PR too.

PR #149, opened 2015-11-26
Add support for normal field name handling.

samsalisbury:
This adds options to enable working around #148 (that is, it enables struct field names to be left untouched rather than lowercased).

All the original tests still pass unchanged, and there are some new tests to verify the new behaviour. This is a fully backwards-compatible change, so no version bump is required.

Example usage to disable the auto lower-case behaviour:

``` go
yaml.Unmarshal(someBytes, &value, yaml.OPT_NOLOWERCASE)  
```

Example usage to use the original standard behaviour:

``` go
yaml.Unmarshal(someBytes, &value)  
```
## Notes:
- This required removing global state, as the global struct
  cache was not amenable to different options being provided
  accross separate calls.
- This is a non-breaking change, and provides only additional
  YAMLCodec type (to replace global state) and variadic
  Options parameters to the Marshal and Unmarshal funcs.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -187,10 +187,11 @@ func (p *parser) mapping() *node {
 // Decoder, unmarshals a node into a provided value.
 
 type decoder struct {
-	doc     *node
-	aliases map[string]bool
-	mapType reflect.Type
-	terrors []string
+	doc       *node
+	aliases   map[string]bool
+	mapType   reflect.Type
+	terrors   []string
+	yamlCodec *YAMLCodec
 }
 
 var (
@@ -200,8 +201,8 @@ var (
 	ifaceType      = defaultMapType.Elem()
 )
 
-func newDecoder() *decoder {
-	d := &decoder{mapType: defaultMapType}
+func newDecoder(y *YAMLCodec) *decoder {
+	d := &decoder{mapType: defaultMapType, yamlCodec: y}
 	d.aliases = make(map[string]bool)
 	return d
 }
@@ -601,7 +602,7 @@ func (d *decoder) mappingSlice(n *node, out reflect.Value) (good bool) {
 }
 
 func (d *decoder) mappingStruct(n *node, out reflect.Value) (good bool) {
-	sinfo, err := getStructInfo(out.Type())
+	sinfo, err := d.yamlCodec.getStructInfo(out.Type())
 	if err != nil {
 		panic(err)
 	}
--- a/decode_test.go
+++ b/decode_test.go
@@ -2,13 +2,14 @@ package yaml_test
 
 import (
 	"errors"
-	. "gopkg.in/check.v1"
-	"gopkg.in/yaml.v2"
 	"math"
 	"net"
 	"reflect"
 	"strings"
 	"time"
+
+	. "gopkg.in/check.v1"
+	"gopkg.in/yaml.v2"
 )
 
 var unmarshalIntTest = 123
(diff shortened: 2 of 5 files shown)


niemeyer:
This is definitely a breaking change, as it's modifying the prototype of a public function. That said, it's not too far from what I want to do.

We need to implement a Encoder/Decoder types that have a method (not a flag list) to enable the standard behavior. The global marshaling/unmarshaling functions simply call an internal global that contains the default encoder/decoder, and their prototype should not change to prevent breaking people's code out there.

We do want to keep the type metadata cache global as well, as the list of types is very limited, and it makes no sense to be paying the cost to recompute that metadata all the time.

Also, please note we need a contributor agreement signed to accept patches: http://www.ubuntu.com/legal/contributors - it's a straightforward non-draconian agreement, easy to sign online, but do read it through to make sure you're comfortable with the clauses there.

I'm closing this for the time being, but please do reopen it if you sort out the above details.

niemeyer:
Ah, one side note if you decide to pursue it: please have a look at the interface of Decoder/Encoder in encoding/json. We want to mimic it, for the same reason you want to mimic its field-naming behavior.

samsalisbury:
@niemeyer OK thanks for the feedback. This is a non-breaking change for the standard use-case (i.e. calling the Marshal and Unmarshal methods directly). I know if you pass those methods around they have a different signature, so fair enough it's not completely safe.

I don't agree with using global variables for the struct cache, unless you also index them by the passed-in options, which IMO is a more fiddly things to manage. (I actually tried this whilst working on the PR, but it was messy.) Alternatively, there could be a _global YAMLCodec_, with a new one only being created for different option sets. With no options being passed in, it would be the same as having a global cache.

WRT the Encoder/Decoder interfaces in `encoding/json` I totally agree it would be great to get this library following similar/the same interfaces, it was on my list for the next PR... Unfortunately I don't have time to contribute that right now, but I will eventually if no one gets there first :)

WRT versioning, I think the cleanest thing to do would probably be to start work on a v3 that has `encoding/json` style behaviour as standard, and provides options for going back to the behaviour exhibited in v2. I'd be very happy to contribute towards that.

So, I'm going to use my fork in the short term, as I really need that behaviour for a project I'm working on, but I'll revisit this when I get a chance (likely in a few weeks' time).

samsalisbury:
Out of interest @niemeyer what is your workflow when using imported from `gopkg.in` that are hosted in GitHub? I.e. how do you use the correct import paths locally?

I have been symlinking `$GOPATH/src/github.com/blah/yaml` dir into `$GOPATH/src/gopkg.in/blah/yaml.v2` but it doesn't feel like a very good way to do it for some reason... Any tips much appreciated :)

samsalisbury:
@niemeyer I've signed the contributor agreement, I named you as the contact I hope that's OK. What do you think about starting on a v3, with `encoding/json` style behaviour as default, and options to enable field name transformations? I think that would follow the principle of least surprise for new users of the project...

niemeyer:
Thanks for signing, Sam, and looking forward to your contributions.

> Alternatively, there could be a global YAMLCodec

Yes, that's what I had in mind too.

> WRT versioning, I think the cleanest thing to do would probably be to start work on a v3

First thing is to get these features into v2 itself. There's no benefit in breaking out right now and making these improvements incompatible from the get go.

> Unfortunately I don't have time to contribute that right now, but I will eventually if no one gets there first :)

Understood. I'll close this PR for the time being then.

> Out of interest @niemeyer what is your workflow when using imported from gopkg.in

Nothing fancy. I just have my branch inside the proper gopkg.in location in $GOPATH.

Thanks for your collaboration, Sam.

PR #152, opened 2015-12-22
make compatible with Go 1.6

mwhudson:
As documented at http://tip.golang.org/doc/go1.6#reflect, the rules around
embedded unexported struct types containing exported fields are changing in Go
1.6. The same release notes explain how to change existing code, so do that.


mwhudson:
Ping.

niemeyer:
Pong. Sorry, need to do a full pass on go-yaml. Will try to get to it tomorrow.

PR #155, opened 2016-01-04
Added functionality to call a BeforeUnmarshalYAML function...

blaubaer:
…direct before start of the unmarshalling of an object. Perfect for set of default values.


blaubaer:
Out of date and not easily mergeable. Create a new one.

PR #162, opened 2016-03-05
Make parser API public

nati:
This new function returns Node struct for usecase which needs
line position of YAML objects for better error message on semantic error
in appliaction use.
This PR also makes Node sturct and related constant public.


niemeyer:
Something similar will come in v3.

PR #163, opened 2016-03-08
Enable stream Decode/Encode

jeffjen:
The pull request enables user to submit an Reader/Writer to process unbounded YAML stream.

Usage example

``` go
// READ from huge yml stream file
r, err := os.Open("thing.yml")
if err != nil {
    panic(err)
}
defer r.Close()

decoder = yaml.NewDecoder(r)
defer decoder.Close()

var thing struct{}
for decoder.Decode(&thing) == nil {
    fmt.Printfln("%#v", stuff)
}
```

``` go
// WRITE to output stream
w, err := os.Create("stuff.yml")
if err != nil {
    panic(err)
}
defer w.Close()

encoder := yaml.NewEncoder(w)
defer encoder.Close()

var stuff struct{}
for encoder.Encode(stuff) == nil {
    // Change date or Adjust??
}

```


niemeyer:
Thanks Jeff.

Can you please have a look at the contributor agreement here and sign if you're comfortable:

It's a non-draconian agreement which is hopefully okay, and may be signed online.

Then, we'll need a few tests and docs, and perhaps some simple tweaks. But the direction is good.

marcosnils:
Any updates on this?. Looking forward for this functionality. @niemeyer

jeffjen:
Need help with "contributor agreement".  I don't know if I am clear on that front.

niemeyer:
@jeffjen Can you please have a look at:

https://www.ubuntu.com/legal/contributors

niemeyer:
The functionality and the API look reasonable. Would be good to have some conventional comments on the public pieces.

guilhem:
Gentle ping?

yamlstream.go
```diff
@@ -0,0 +1,64 @@
...
+func (enc *Encoder) Encode(in interface{}) (err error) {
+	defer handleErr(&err)
+	enc.e.begin()
+	enc.e.marshal("", reflect.ValueOf(in))
+	enc.e.end()
+	return
+}
+
+func (enc *Encoder) Close() {
+	enc.e.finish()
```
rogpeppe, line 56: I think that this needs a defer handleErr otherwise it can panic if there's an error when the final bytes are written to the stream.

yamlstream.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func NewDecoder(r io.Reader) *Decoder {
+	return &Decoder{
+		d: newDecoder(),
+		p: newFileParser(r),
+	}
+}
+
+type Encoder struct {
```
rogpeppe, line 43: doc comment please.

yamlstream.go
```diff
@@ -0,0 +1,64 @@
...
+	e *encoder
+}
+
+func (enc *Encoder) Encode(in interface{}) (err error) {
```
rogpeppe, line 47: doc comment please.

yamlstream.go
```diff
@@ -0,0 +1,64 @@
...
+		return &TypeError{dec.d.terrors}
+	}
+	return nil
+}
+
+func (dec *Decoder) Close() {
+	dec.p.destroy()
+}
+
+func NewDecoder(r io.Reader) *Decoder {
```
rogpeppe, line 36: doc comment please.

yamlstream.go
```diff
@@ -0,0 +1,64 @@
...
+		}
+		dec.d.unmarshal(node, v)
+	}
+	if len(dec.d.terrors) > 0 {
+		return &TypeError{dec.d.terrors}
+	}
+	return nil
+}
+
+func (dec *Decoder) Close() {
```
rogpeppe, line 32: I don't think this method is needed - all the destroy method does is zero some memory. I think we can leave the garbage collector to do that.

yamlstream.go
```diff
@@ -0,0 +1,64 @@
...
+	"io"
+	"reflect"
+)
+
+type Decoder struct {
+	d *decoder
+	p *parser
+}
+
+func (dec *Decoder) Decode(out interface{}) (err error) {
```
rogpeppe, line 13: doc comment please.

yamlstream.go
```diff
@@ -0,0 +1,64 @@
+package yaml
+
+import (
+	"io"
+	"reflect"
+)
+
+type Decoder struct {
```
rogpeppe, line 8: doc comment please.

yamlstream.go
```diff
@@ -0,0 +1,64 @@
...
+	enc.e.end()
+	return
+}
+
+func (enc *Encoder) Close() {
+	enc.e.finish()
+	enc.e.destroy()
+}
+
+func NewEncoder(w io.Writer) *Encoder {
```
rogpeppe, line 60: doc comment please.

yamlstream_test.go
```diff
@@ -0,0 +1,106 @@
...
+	err := decoder.Decode(value)
+	if _, ok := err.(*yaml.TypeError); !ok {
+		if err != io.EOF {
+			c.Assert(err, IsNil)
+		}
+	}
+	if t.Kind() == reflect.String {
+		c.Assert(*value.(*string), Equals, v.Interface())
+	} else {
+		c.Assert(value, DeepEquals, v.Interface())
```
rogpeppe, line 71: We should be able to use DeepEquals even when the value is a string.

decode.go
```diff
@@ -57,6 +58,22 @@ func newParser(b []byte) *parser {
...
 
+func newFileParser(r io.Reader) *parser {
+	p := parser{}
+	if !yaml_parser_initialize(&p.parser) {
+		panic("failed to initialize YAML emitter")
+	}
+
+	yaml_parser_set_input_file(&p.parser, r)
+
+	p.skip()
```
rogpeppe, line 69: I'm not sure we should be calling skip here -  it means that NewDecoder can fail, and I don't think it's a great idea to make that return an error (it's nice if the YAML API mirrors the encoding/json API).

encode.go
```diff
@@ -30,6 +31,26 @@ func newEncoder() (e *encoder) {
 	return e
 }
 
+func newFileEncoder(w io.Writer) (e *encoder) {
+	e = &encoder{}
+	e.must(yaml_emitter_initialize(&e.emitter))
+	yaml_emitter_set_output_file(&e.emitter, w)
+	yaml_emitter_set_unicode(&e.emitter, true)
+	e.must(yaml_stream_start_event_initialize(&e.event, yaml_UTF8_ENCODING))
```
rogpeppe, line 39: A similar issue applies here to the one in newFileParser. I don't think this should actually write anything until a value is encoded.

yamlstream_test.go
```diff
@@ -0,0 +1,106 @@
+package yaml_test
```
rogpeppe, line 1: We'd need some Encoder tests too, please.

rogpeppe (changes requested):
This looks like a good start, thanks, but there are some issues that need resolving before this can land, I think.

im-kulikov:
https://github.com/go-yaml/yaml/pull/288 commits + would fix some issues

niemeyer:
Implemented in #301 and #306.

PR #170, opened 2016-04-21
Support for go-like field name handling (100% v2 backwards compat)

samsalisbury:
This is a follow on from #149, adding 2 extra methods `MarshalWithOptions` and `UnmarshalWithOptions` (both accepting options to control field name handling) and leaving `Marshal` and `Unmarshal` with the same signature as v2 has currently.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -187,10 +187,11 @@ func (p *parser) mapping() *node {
 // Decoder, unmarshals a node into a provided value.
 
 type decoder struct {
-	doc     *node
-	aliases map[string]bool
-	mapType reflect.Type
-	terrors []string
+	doc       *node
+	aliases   map[string]bool
+	mapType   reflect.Type
+	terrors   []string
+	yamlCodec *YAMLCodec
 }
 
 var (
@@ -200,8 +201,8 @@ var (
 	ifaceType      = defaultMapType.Elem()
 )
 
-func newDecoder() *decoder {
-	d := &decoder{mapType: defaultMapType}
+func newDecoder(y *YAMLCodec) *decoder {
+	d := &decoder{mapType: defaultMapType, yamlCodec: y}
 	d.aliases = make(map[string]bool)
 	return d
 }
@@ -601,7 +602,7 @@ func (d *decoder) mappingSlice(n *node, out reflect.Value) (good bool) {
 }
 
 func (d *decoder) mappingStruct(n *node, out reflect.Value) (good bool) {
-	sinfo, err := getStructInfo(out.Type())
+	sinfo, err := d.yamlCodec.getStructInfo(out.Type())
 	if err != nil {
 		panic(err)
 	}
--- a/decode_test.go
+++ b/decode_test.go
@@ -966,9 +966,9 @@ func (s *S) TestUnmarshalUppercaseFieldNamesNormal(c *C) {
 
 }
 
-func (s *S) TestUnmarshalUppercaseFieldNamesWithoutLowercase(c *C) {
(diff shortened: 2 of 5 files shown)


kindlyseth:
Current v2 is unable to parse the files I already have directly into the 3rd party data structures they need to be loaded into because it only tries the case converted names.

I could convert the files, but simply converting the field names to lower case would be the wrong conversion. Converting to snake case would be correct, although not better in this case, but then the files still wouldn't load. I'll end up using an intermediate data structure, but that's effort I wouldn't have to spend if this pull request had been merged.

It's been 5 months since this PR was submitted and there's no record of feedback, so it doesn't appear that sufficient resources are being applied. If someone's time needs to be sponsored, I might be able to apply resources. If the problem is lack of recruiting trusted people, I'm less able.

samsalisbury:
I would obviously really like to get this change made as well, I've been using my own fork for ages because of this. @niemeyer any reason not to merge this?

kindlyseth:
I ended up going the same route, creating https://github.com/kindlyops/yaml which I expect to make track the v2 branch but with the one line change to match field names exactly applied.

niemeyer:
We've discussed this multiple times in other PRs and issues. We can't have a global option like this, because it would break other packages that are being imported into the same binary and have different expectations. The proper path here is to have an Encoder and Decoder which may be tuned. We already have in progress in other PRs, and I'll move it forward once I get the unrelated work I'm doing out of the door.

samsalisbury:
@niemeyer this does not add any global options, just 3 new package-level functions `UnmarshalWithOptions`, `MarshalWithOptions` and `NewYAMLCodec`. It also adds a couple of public constants, but code not using these functions or constants will still work. The original public package functions `Marshal` and `Unmarshal` are unchanged in this PR, so _this PR would not break any usages of v2 at all_. We had a conversation about my original change in #149 which could have broken usages, but this PR addresses those issues so that no usages can be broken.

niemeyer:
Sorry, yes, you are right. I misread the API as setting global options instead of unmarshaling in place.

The idea which seems best is still the one we discussed in that thread, though, which differs from what was presented as the follow up:

> We need to implement a Encoder/Decoder types that have a method (not a flag list) to enable the standard behavior. The global marshaling/unmarshaling functions simply call an internal global that contains the default encoder/decoder, and their prototype should not change to prevent breaking people's code out there.

samsalisbury:
So, I should change "YAMLCodec" to 2 types "Encoder" and "Decoder", that's easy enough to do, although I'm not sure the benefit. What should the function/method be called that "enables the standard behaviour"? Do you mean a global function called e.g. `yaml.NoLowercase()` that would affect all encoding/decoding?

It seems best not to use globals for that setting, since I may have a program that needs to use both behaviours. The idea of the YAMLCodec is to encapsulate YAML formatting options that enable a complete roundtrip of YAML following whichever convention is decided per codec (I am assuming that usually you want to read and write in the same format.) These codecs can be used as singletons, so you only need one codec of each kind per execution. I am happy to add the global setting as well if that will get the PR merged, and I am also happy to drop the flags if they are a problem (to me they are descriptive, but that's just an opinion).

niemeyer:
> Do you mean a global function called e.g. yaml.NoLowercase() that would affect all encoding/decoding?

No, per the quoted sentence: _"We need to implement a Encoder/Decoder types that have a method (not a flag list) to enable the standard behavior."_

Please have a look at encoding/json, for example. Perhaps `UseFieldNames` for the method name.

PR #171, opened 2016-04-26
Tighten restrictions on float decoding

alexjh:
The Golang checks for floats are more permissive than the YAML spec, causing problems when reading in commit hashes (strings) that are occasionally confused with floats. This change also matches the Ruby and Python implementations' behaviour.


alexjh:
~~I'm looking into the CLA agreements that HPE Cloud has with Canonical, will update when I have more info.~~

Signed the individual agreement as we don't have an org-wide one set up.

alexjh:
@niemeyer Just wanted to ping you on this, I realized that I edited the comment but that may not have generated a new notification.

aarondl:
Could really use this. It prevents yaml generated from other languages from being consumed by Go.

alexjh:
@niemeyer Since the license changed, should I close and resubmit this PR? I'm fine with my change being Apache 2 licensed, so I can also leave it as is.

niemeyer:
Thanks, and sorry for the long delay.

PR #173, opened 2016-05-03
Remove unreachable code

ryboe:
Removes three cases of unreachable code found by `go vet`. There are still `go vet` warnings coming from some of the test files, but those are less important to fix. Most vendoring tools do not fetch test files. Anybody using `gvt`, `govendor`, etc. will be able to run `go vet ./...` on their own projects and not see any warnings from `go-yaml`.


ryboe:
Fixed in #181, with additional fixes for the Travis build.

PR #178, opened 2016-06-06
Add package option for mimicking encoding/json field name case handling

parsley42:
I read the other pull request for similar functionality, but didn't want to implement as part of a Decoder/Encoder because it wouldn't give the behavior I wanted when using standard Marshal/Unmarshal. This non-breaking change adds a package option to preserve the case of
field names when marshaling/unmarshaling.

If this approach is acceptable, I'll be glad to get a launchpad ID and fill out the form, but didn't want to bother without getting feedback first. Thanks!

### DIFF
--- a/yaml.go
+++ b/yaml.go
@@ -145,6 +145,14 @@ func Marshal(in interface{}) (out []byte, err error) {
 	return
 }
 
+// Individual apps can use SetPreserveCase(true) to mimic the encoding/json
+// behavior that preserves the case of struct fields.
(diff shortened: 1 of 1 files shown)


rogpeppe:
I'm afraid I don't think that having a global variable that modifies YAML unmarshaling behaviour is a good idea. It means that if you have several packages that use YAML in different ways, they can step on each others' toes. In general, package-level mutable variables should be considered a bad idea.

PR #181, opened 2016-06-08
remove unreachable code to avoid 'go vet' warnings

dmitris:
same as https://github.com/go-yaml/yaml/pull/173 with additional change in .travis.yaml to fix the build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,5 @@ go:
     - tip
 
 go_import_path: gopkg.in/yaml.v2
+
+script: go get -t -v ./...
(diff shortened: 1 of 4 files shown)


mogthesprog:
Is it possible to get this merged?

niemeyer:
@mogthesprog Why do we need the additional go get call in travis?

mogthesprog:
I hadn't noticed that. I'm pretty sure it's not needed at all. I can raise another PR without that change in it?

niemeyer:
@mogthesprog Yes.. #173 was exactly that, but unfortunately it got removed after this PR.

Happy to merge if you can redo it.

mogthesprog:
Sure, no worries. Bear with me a sec..

niemeyer:
Sorted out in #249.

PR #185, opened 2016-07-25
Preserve JSON Codec Compatibility

akutz:
This patch adds the function `PreserveJSONCodecCompatibility` in order to provide the ability to preserve data unmarshaled using this package and marshal it back to JSON using the Golang StdLib JSON library.

The Go StdLib JSON marshaller cannot encode maps with keys other than of type string. Therefore, maps created by this package's decoder must use a string as the key-type in order for the same maps to be marshaled into JSON at a later date.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -194,11 +194,33 @@ type decoder struct {
 }
 
 var (
-	mapItemType    = reflect.TypeOf(MapItem{})
-	durationType   = reflect.TypeOf(time.Duration(0))
-	defaultMapType = reflect.TypeOf(map[interface{}]interface{}{})
-	ifaceType      = defaultMapType.Elem()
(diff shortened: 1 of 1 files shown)


akutz:
Hi,

I was wondering if there are any thoughts on this? Thanks!

akutz:
Hi,

I was wondering if there are any thoughts on this? Thanks!

akutz:
Hi there. I suppose this is just a dead repository then.

niemeyer:
It's far from dead, but it's just not a priority for anyone to respond to such feature requests, so it may take a while, sorry about that.

The particular feature you're asking for will be supported in v3 by simply unmarshaling map[string]interface{}. It will preserve the choice for nested maps too. We cannot do that in v2 because it's incompatible.

Issue #192, opened 2016-08-18
Marshaler incorrectly marshals date strings


autrilla:
The marshaler should quote strings that have YAML's date format. For example, the go string "2006-01-02T15:04:05Z" should be marshaled as '2006-01-02T15:04:05Z', with the quotes, but is currently marshaled without quotes.

SSCCE:

```
package main

import (
    "fmt"
    "gopkg.in/yaml.v2"
)

func main() {
    m := make(map[string]string)
    m["foo"] = "2006-01-02T15:04:05Z"
    out, _ := yaml.Marshal(m)
    fmt.Println(string(out))
}
```

Should output:

`foo: '2006-01-02T15:04:05Z'`

But instead outputs

`foo: 2006-01-02T15:04:05Z`

niemeyer:
Thanks for this.

rogpeppe:
Given that YAML directly supports unquoted RFC3339 times, I think this is OK.
The marshaled YAML correctly unmarshals AFAICS.
Please re-open if this is still not OK.

autrilla:
Yes, YAML supports unquoted times, but a Go string formatted like a time (e.g. `s := "2006-01-02T15:04:05Z"`) will be output as a YAML date (unquoted), not as a string. I don't think this silent type conversion is OK.

rogpeppe:
@autrilla Thanks, I'd gone through this issue too quickly and missed that it wasn't a time being printed. Will fix, thanks.

autrilla:
I opened a PR for this back in the day on https://github.com/go-yaml/yaml/pull/213 , it seems there's some merge conflicts now though. I'll see if I can get them fixed.

niemeyer:
This case is already fixed.

PR #193, opened 2016-08-18
Correctly quote date-like strings (issue #192)

autrilla:


encode.go
```diff
@@ -233,10 +236,23 @@ func isBase60Float(s string) (result bool) {
...
 // From http://yaml.org/type/float.html, except the regular expression there
 // is bogus. In practice parsers do not enforce the "\.[0-9_]*" suffix.
 var base60float = regexp.MustCompile(`^[-+]?[0-9][0-9_]*(?::[0-5]?[0-9])+(?:\.[0-9_]*)?$`)
 
+// From http://yaml.org/type/timestamp.html
+var date = regexp.MustCompile(`[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]|[0-9][0-9][0-9][0-9]-[0-9][0-9]?-[0-9][0-9]?([Tt]|[ \t]+)[0-9][0-9]?:[0-9][0-9]:[0-9][0-9](\.[0-9]*)?(([ \t]*)Z|[-+][0-9][0-9]?(:[0-9][0-9])?)?`)
+
+func (e *encoder) timev(tag string, in reflect.Value) {
+	m := in.Interface().(encoding.TextMarshaler)
+	s, _ := m.MarshalText()
```
tv42: Handle your errors.

encode.go
```diff
@@ -233,10 +236,23 @@ func isBase60Float(s string) (result bool) {
...
+
 // From http://yaml.org/type/float.html, except the regular expression there
 // is bogus. In practice parsers do not enforce the "\.[0-9_]*" suffix.
 var base60float = regexp.MustCompile(`^[-+]?[0-9][0-9_]*(?::[0-5]?[0-9])+(?:\.[0-9_]*)?$`)
 
+// From http://yaml.org/type/timestamp.html
+var date = regexp.MustCompile(`[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]|[0-9][0-9][0-9][0-9]-[0-9][0-9]?-[0-9][0-9]?([Tt]|[ \t]+)[0-9][0-9]?:[0-9][0-9]:[0-9][0-9](\.[0-9]*)?(([ \t]*)Z|[-+][0-9][0-9]?(:[0-9][0-9])?)?`)
+
+func (e *encoder) timev(tag string, in reflect.Value) {
+	m := in.Interface().(encoding.TextMarshaler)
```
tv42: This repeats the type assertion from earlier, probably better to make the argument of the right type and just let parent type assert.

autrilla: I don't handle the case of in.Interface() not being a TextMarshaler because that's checked earlier on. I guess in doesn't need to be reflect.Value, but I left it that way because the rest of the functions are also like that.

Issue #194, opened 2016-08-23
Unstable marshalled output for map with keys containing numbers


prashantv:
When comparing map keys, if both keys contain a number starting with the same digit, then there's a bug in the comparison.

This test reproduces the multiple orders:

``` go
package main

import (
    "fmt"
    "os"

    yaml "gopkg.in/yaml.v2"
)

func main() {
    vals := map[string]int{"a/100": 1, "a/17": 2, "a/20": 3}
    generated := make(map[string]struct{})
    for i := 0; i < 100; i++ {
        contents, err := yaml.Marshal(vals)
        if err != nil {
            panic(err)
        }
        generated[string(contents)] = struct{}{}
    }

    if len(generated) != 1 {
        fmt.Println("Generated multiple outputs")
        for k := range generated {
            fmt.Println(k)
            fmt.Println("----")
        }
        os.Exit(-1)
    }
}
```

The output is:

```
Generated multiple outputs
a/17: 2
a/20: 3
a/100: 1

----
a/20: 3
a/100: 1
a/17: 2

----
a/100: 1
a/17: 2
a/20: 3

----
exit status 255
```

The issue is when skipping runes if they are the same. This causes issues when skipping digits, since comparing `a/100` to `a/17` leads to the `1` being skipped, and then the values are compared as `00` vs `7`. This leads to `a/100` sorting to less than `a/17`. However, if the values were compared with `a/20` instead, the values are sorted in the correct order.

benfleis:
+1

Can this get some love?

prashantv:
@niemeyer Would you be able to take a look at this?

vinzenz:
Actually this is not a problem with numbers. This is just this: 

> When iterating over a map with a range loop, the iteration order is not specified and is not guaranteed to be the same from one iteration to the next. Since Go 1 the runtime randomizes map iteration order, as programmers relied on the stable iteration order of the previous implementation. If you require a stable iteration order you must maintain a separate data structure that specifies that order. This example uses a separate sorted slice of keys to print a map[int]string in key order:

See: https://blog.golang.org/go-maps-in-action # Iteration order

prashantv:
@vinzenz The yaml marshaller produces a stable output by iterating over keys in sorted order, if you saw the code here:
https://github.com/prashantv/yaml/blob/v2/sorter.go

[...]

PR #195, opened 2016-08-23
Fix unstable map ordering for specific keys with numbers

prashantv:
When comparing runes, don't skip equal runes if they are digits. This avoids issues when comparing values like `100` vs `11` since dropping the `1` causes a comparison of `00` vs `1`.

Fixes https://github.com/go-yaml/yaml/issues/194

### DIFF
--- a/sorter.go
+++ b/sorter.go
@@ -39,7 +39,9 @@ func (l keyList) Less(i, j int) bool {
 	ar, br := []rune(a.String()), []rune(b.String())
 	for i := 0; i < len(ar) && i < len(br); i++ {
 		if ar[i] == br[i] {
-			continue
+			if !(unicode.IsDigit(ar[i]) && unicode.IsDigit(br[i])) {
+				continue
+			}
 		}
 		al := unicode.IsLetter(ar[i])
 		bl := unicode.IsLetter(br[i])
@@ -62,6 +64,9 @@ func (l keyList) Less(i, j int) bool {
 		}
 		if ai != bi {
 			return ai < bi
+		}
+		if ar[i] == br[i] {
+			continue
 		}
 		return ar[i] < br[i]
 	}
--- a/encode_test.go
+++ b/encode_test.go
@@ -7,10 +7,11 @@ import (
 	"strings"
 	"time"
 
-	. "gopkg.in/check.v1"
-	"gopkg.in/yaml.v2"
 	"net"
 	"os"
+
+	. "gopkg.in/check.v1"
+	"gopkg.in/yaml.v2"
 )
 
 var marshalIntTest = 123
@@ -456,7 +457,8 @@ func (s *S) TestSortedOutput(c *C) {
 		"2",
 		"a!10",
 		"a/2",
-		"a/10",
+		"a/11",
+		"a/100",
 		"a~10",
 		"ab/1",
 		"b/1",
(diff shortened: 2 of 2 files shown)


prashantv:
Hi @niemeyer, would you be able to look at this change? I'd love to get this merged in and avoid maintaining a fork!

geofffranks:
I would love to see this merged as well

sorter.go
```diff
@@ -63,6 +65,9 @@ func (l keyList) Less(i, j int) bool {
 		if ai != bi {
 			return ai < bi
 		}
+		if ar[i] == br[i] {
+			continue
```
rogpeppe, line 69: I suspect this may not be quite right. The original code always returned after comparing numbers, but this now means that we can carry on instead. At the least, this introduces quadratic behaviour which might be a problem for long keys with many decimal digits in. I'll need to think some more about this.

rogpeppe (commented):
Thanks for this, but I'll need to think some more before submitting it. The keyList.Less method looks problematic already, in fact, as it looks like it's assuming that unicode.IsDigit implies that the rune value is between ASCII '0' and '9'.

sorter.go
aviadl: @rogpeppe is there a chance that this would be fixed anytime soon, it really hurts us on some scenarios

aviadl: Can i offer to simple do a string comparison on the last case:
So the less function would look like this:
```
func (l keyList) Less(i, j int) bool {
	a := l[i]
	b := l[j]
	ak := a.Kind()
	bk := b.Kind()
	for (ak == reflect.Interface || ak == reflect.Ptr) && !a.IsNil() {
		a = a.Elem()
		ak = a.Kind()
	}
	for (bk == reflect.Interface || bk == reflect.Ptr) && !b.IsNil() {
		b = b.Elem()
		bk = b.Kind()
	}
	af, aok := keyFloat(a)
	bf, bok := keyFloat(b)
	if aok && bok {
		if af != bf {
			return af < bf
		}
		if ak != bk {
			return ak < bk
		}
		return numLess(a, b)
	}
	if ak != reflect.String || bk != reflect.String {
		return ak < bk
	}
	ar, br := a.String(), b.String()
	return ar < br
}
```

My mine use case is that the output would be consistent.
I don't really care about the order of map, as long as it is consitent

jfmyers9:
Hey @rogpeppe,

We are also running into issues with the non-stable ordering in maps. Is there a timeline for when this or another fix may be merged?

Best,

@jfmyers9 && @tjvman

niemeyer:
@jfmyers9 I'll fix it this week, somehow.

cppforlife:
@niemeyer any updates?

niemeyer:
This is fixed in https://github.com/go-yaml/yaml/commit/dcd83b31fd165d8cc8677fce58f889dca3e06f35, without significant changes to the algorithm overall behavior.

Please let me know how it goes.

HeavyWombat:
Just in case someone comes here like me and thinks this is completely solved, I can still reproduce unstable map ordering with a use case I have. Run the following code a couple of times and you will observe different orders:
```go
package main

import (
	"fmt"

	"gopkg.in/yaml.v2"
)

const example = `
properties:
  cc:
    quota_definitions:
      q256MB:
        non_basic_services_allowed: true
      q2GB:
        non_basic_services_allowed: true
      q4GB:
        non_basic_services_allowed: true
`

func main() {
	var data interface{}
	if err := yaml.Unmarshal([]byte(example), &data); err == nil {
		if out, err := yaml.Marshal(data); err == nil {
			fmt.Println(string(out))
		}
	}
}

```

PR #189, opened 2016-08-16
new license in the README file

chillum:
Hi,

Recently go-yaml updated license to Apache v2.

Unfortunately, that was not reflected in the README file.

PR #199, opened 2016-09-12
Updating license reference in README

brosander:
@niemeyer I noticed the LICENSE file is now Apache, here's updated text for README.


brosander:
@niemeyer Looks like I missed the boat, #189 already has same change 😄

PR #201, opened 2016-09-19
don't allow invalid durations

### LINKED ISSUE #200: go-yaml accepts invalid durations
When you call ParseDuration with an invalid duration (e.g. `15`) it errors: https://play.golang.org/p/WAq-o3NT3D

However, if you use go-yaml and try to marshal an invalid duration such as `15` into a `time.Duration`, it accepts it and treats it as a number of nanoseconds.

test.yaml

``` yaml
thing: 15
```

test.go

``` go
package main

import (
        "io/ioutil"
        "log"
        "time"

        yaml "gopkg.in/yaml.v2"
)

type Spec struct {
        Thing time.Duration `yaml:"thing"`
}

func main() {
        data, err := ioutil.ReadFile("./test.yaml")
        if err != nil {
            panic(err)
        }

        var s Spec
        if err := yaml.Unmarshal(data, &s); err != nil {
            panic(err)
        }

        log.Printf("%#v", s)
}
```

output

```
2016/09/19 13:48:07 main.Spec{Thing:15}
```

azylman:
This fixes https://github.com/go-yaml/yaml/issues/200

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -379,24 +379,28 @@ func (d *decoder) scalar(n *node, out reflect.Value) (good bool) {
 		}
 		good = true
 	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		// The only valid types for durations are strings, nothing else
+		// can have a parseable unit
+		isDuration := out.Type() == durationType
+
 		switch resolved := resolved.(type) {
 		case int:
-			if !out.OverflowInt(int64(resolved)) {
+			if !isDuration && !out.OverflowInt(int64(resolved)) {
 				out.SetInt(int64(resolved))
 				good = true
 			}
 		case int64:
-			if !out.OverflowInt(resolved) {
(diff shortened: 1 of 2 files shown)


m-masataka:
@niemeyer @azylman Why don't you merge this commit?
It's critical issue for another product that include ``go-yaml/yaml``.

niemeyer:
This a backwards incompatible change, and not clearly a bug in the sense that a time.Duration is in fact a number and not a string. So the main question is whether we want to do it at all. I can see it going both ways at the moment.

m-masataka:
@niemeyer
Thanks for reply.
I make sure your answer and changed my mind.
It's not go-yaml's  job.
However, it's couse some bugs. Does go-yaml has any documents? If not, we should generate warning for omitting the unit of time.
If you need, I implement the warning function or add the documentation.

niemeyer:
This issue will be addressed in v3. Will close #200 once v3 is available.

PR #204, opened 2016-09-24
custom tags support

jdamick:
i'd like to be able handle custom yaml tags; similar support is found in other language yaml implementations.  So this changes adds support to register a tag unmarshaler.  Please let me know if this is something you'd like to handle a different way & i'd be happy to modify it.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -194,16 +194,21 @@ type decoder struct {
 }
 
 var (
-	mapItemType    = reflect.TypeOf(MapItem{})
-	durationType   = reflect.TypeOf(time.Duration(0))
-	defaultMapType = reflect.TypeOf(map[interface{}]interface{}{})
-	ifaceType      = defaultMapType.Elem()
+	mapItemType     = reflect.TypeOf(MapItem{})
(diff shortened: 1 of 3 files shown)


technosophos:
I'd like this feature as well.

tomnz:
We're trying to support custom parsing into decimals with go-yaml. Currently we have to special case it (look for a particular structure e.g. `{ decimal: "0.123" }` in the YML), but it would be much cleaner to parse it as `!dec "0.123"` instead. Looks like this diff would work great! It's a shame we can't use custom YAML tags currently.

jdamick:
@niemeyer any chance we can get this feature added?

rogpeppe:
I'm not convinced that custom tag unmarshalers should be globally registered. Also, this only covers tagged strings, not tagged sequences or mappings.

lox:
Completely agree @rogpeppe, seems like this will probably be superseded by https://github.com/go-yaml/yaml/pull/332

niemeyer:
This will be doable in v3 by unmarshalling into the intermediate representation and then using it to decode as desired.

PR #211, opened 2016-10-13
Added functionality to call a BeforeUnmarshalYAML function direct before start of the unmarshalling of an object.

blaubaer:
... Perfect for set of default values.


rogpeppe:
ISTM that it's easy to do this with UnmarshalYAML only: https://play.golang.org/p/lE6xOX-Qei2 so closing as we can do this already. Please reopen if it there's something this can do that UnmarshalYAML can't.

PR #213, opened 2016-10-20
Correctly quote date-like strings

### LINKED ISSUE #192: Marshaler incorrectly marshals date strings

autrilla:
Fixes #192 

(this is a copy of #193, but since then, we've needed to move the fix to another branch)

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -75,6 +75,9 @@ func (e *encoder) marshal(tag string, in reflect.Value) {
 			return
 		}
 		in = reflect.ValueOf(v)
+	} else if _, ok := iface.(time.Time); ok {
+		e.timev(tag, in)
+		return
 	} else if m, ok := iface.(encoding.TextMarshaler); ok {
 		text, err := m.MarshalText()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


autrilla:
I'm not sure why this is broken on tip

vinzenz:
@autrilla It's fixed by #217 - It's a change in the time stdlib

rogpeppe:
Fixed by #310.

autrilla:
@rogpeppe I'm not sure this is fixed, unless I'm misunderstanding things. Running the reproduction case from the issue:

```
cd /tmp && mkdir gotmp && export GOPATH=/tmp/gotmp && go get -u gopkg.in/yaml.v2 && cat <<EOF > main.go && go run main.go
package main

import (
    "fmt"
    "gopkg.in/yaml.v2"
)

func main() {
    m := make(map[string]string)
    m["foo"] = "2006-01-02T15:04:05Z"
    out, _ := yaml.Marshal(m)
    fmt.Println(string(out))
}
EOF
```

Outputs:

```
foo: 2006-01-02T15:04:05Z
```

But if it was fixed, it should have output the date string quoted with single or double quotes.

Edit: Oh, never mind, I see this is on the `devel` branch.

PR #216, opened 2016-10-31
Fix misspell: preceed

come-maiz:
https://en.wiktionary.org/wiki/preceed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,9 @@ go:
     - 1.4
     - 1.5
     - 1.6
+    - 1.7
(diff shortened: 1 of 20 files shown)


niemeyer:
This matches the upstream libyaml code. Please submit it there first so comparing the two is easier.

come-maiz:
fwiw: https://github.com/yaml/libyaml/commit/37c92e06f887c85d39633ae2a055381ebe1c3d63

niemeyer:
Thanks!

PR #217, opened 2016-11-01
Fix decode test for Go 1.8

dsnet:
An upstream change (http://golang.org/cl/31144) alters the time package
to be more consistent about the representation of UTC location.
We adjust the time.Time used in the test to work on both 1.7 and 1.8.


dsnet:
\cc @niemeyer, can we get this merged?
Thanks.

niemeyer:
Thanks!

Issue #132, opened 2015-10-07
Support (un)marshaling of comments 


DirectXMan12:
Reopening from #129

It would be useful if go-yaml supported outputting comments when marshaling an object into YAML.

The relevant bits of the YAML specification:
- Intro to comments: http://yaml.org/spec/1.2/spec.html#id2767100
- Comment specification: http://yaml.org/spec/1.2/spec.html#id2780069
- See also http://yaml.org/spec/1.2/spec.html#id2762313

Comments can be useful to provide human-readable information about the contents of a YAML document without affecting the actual content of the YAML itself.  For instance, comments could be used to provide documentation about fields in generated configuration information, or to point out errors in the configuration (e.g. "key A may only have values X, Y, or Z").

DirectXMan12:
> We can certainly support comment manipulation, but we'll need to design this a little bit better first.
> 
> Namingly, we need to ignore the internal implementation completely and look from the perspective of someone using the feature both to consume comments and to produce them back into the expected locations.

According to the YAML 1.2 specification, comments are not supposed to live anywhere but the presentation form (see http://yaml.org/spec/1.2/spec.html#id2762140 , http://yaml.org/spec/1.2/spec.html#id2767100 , http://yaml.org/spec/1.2/spec.html#id2766150), and are not supposed to be "attached" to any particular node.

Since comments are not supposed to survive into the unmarshaled data structure, and other presentation details (e.g. style) don't survive a round trip of unmarshaling into `interface{}` and then remarshaling, IMO we don't need to worry about presenting deserialized comments to the user or having the "survive" a round trip.

Currently, it appears that go-yaml supports three ways of influencing the marshaling process:
- Using `MarshalYAML` (and/or `MarshalText`), the user can control the representation stage
- Using struct tags, the user can affect the representation stage (`inline` and `omitempty`) and the presentation stage (`flow`).
- Using `MapItem` and `MapSlice`, the user can affect the serialization phase (specifically the key ordering part)

In my opinion, the ability to have "dynamic" (i.e. not hard-coded) comments is required (e.g. in the case of indicating validation errors, as I described above).  Ergo, while the struct tag method could be an option for adding comments, it cannot be the only option.  If we wish to utilize the existing machinery, this means we are left with two options -- either an extension to `MarshalYAML` that somehow adds in comment comments (this allows for some degree of dynamic comments but makes certain things hard), or attaching comments to `MapItem` values (as I implemented in #129).

[...]

PR #219, opened 2016-11-02
Comment marshaling and unmarshaling support

autrilla:
This introduces support for marshaling and unmarshaling comments.

Some key points:
* Comment unmarshaling is opt-in.
* We don't retain indentation information for comments. Comments are placed at the same indentation level as the previous token.
* The implementation uses the exact same structure as all current YAML tokens (so it's tokenized, an event is pushed, etc).

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -2,8 +2,8 @@ package yaml_test
 
 import (
 	"errors"
-	"go.mozilla.org/yaml.v2"
 	. "gopkg.in/check.v1"
+	"gopkg.in/yaml.v2"
 	"math"
 	"net"
 	"reflect"
--- a/decode_test.go
+++ b/decode_test.go
@@ -189,14 +189,14 @@ var unmarshalTests = []struct {
 
 	// Literal block scalar
 	{
-		"scalar: | # Comment\n\n literal\n\n \ttext\n\n",
-		map[string]string{"scalar": "\nliteral\n\n\ttext\n"},
+		"scalar: |\n literal\n\n \ttext\n\n",
+		map[string]string{"scalar": "literal\n\n\ttext\n"},
 	},
 
 	// Folded block scalar
 	{
-		"scalar: > # Comment\n\n folded\n line\n \n next\n line\n  * one\n  * two\n\n last\n line\n\n",
-		map[string]string{"scalar": "\nfolded line\nnext line\n * one\n * two\n\nlast line\n"},
+		"scalar: >\n folded\n line\n \n next\n line\n  * one\n  * two\n\n last\n line\n\n",
+		map[string]string{"scalar": "folded line\nnext line\n * one\n * two\n\nlast line\n"},
 	},
 
 	// Map inside interface with no type hints.
@@ -706,7 +706,7 @@ func (s *S) TestUnmarshalerPointerField(c *C) {
 		if item.value == nil {
 			c.Assert(obj.Field, IsNil)
 		} else {
-			c.Assert(obj.Field, NotNil, Commentf("Pointer not initialized (%#v)", item.value))
+			c.Assert(obj.Field, NotNil, Commentf("Pointer not initialized (#v)", item.value))
 			c.Assert(obj.Field.value, DeepEquals, item.value)
 		}
 	}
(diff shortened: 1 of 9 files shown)


decode_test.go
```diff
@@ -2,8 +2,8 @@ package yaml_test
 
 import (
 	"errors"
+	"go.mozilla.org/yaml.v2"
 	. "gopkg.in/check.v1"
-	"gopkg.in/yaml.v2"
```
jvehent: these import paths changes need to be reverted

encode_test.go
```diff
@@ -7,8 +7,8 @@ import (
 	"strings"
 	"time"
 
+	"go.mozilla.org/yaml.v2"
 	. "gopkg.in/check.v1"
-	"gopkg.in/yaml.v2"
```
jvehent: these import path changes need to be reverted

jvehent:
Tests are failing due to https://github.com/go-yaml/yaml/pull/223/commits/e09b2a6483d4262d3b5cadc8a6f81d539981fc22, therefore this is blocked on merging #223

autrilla:
The tests now pass, cc @jvehent

buchanae:
I'm missing a comment from the "path" key in the test program below. Notice the comment **do** show up for all other keys (e.g. "server")

```go
package main

import "fmt"
import "github.com/mozilla-services/yaml"

var data = `
# Database
database:
  # Database path
  path: 'path/to/db'
  # Server port
  server: 9090
  # Sub comment
  sub:
    val: foo
`

func main() {
  b := []byte(data)
  unmarshaler := yaml.CommentUnmarshaler{}
  doc := yaml.MapSlice{}
  err := unmarshaler.Unmarshal(b, &doc)
  if err != nil {
    panic(err)
  }

  ob, _ := yaml.Marshal(doc)
  fmt.Println(doc)
  fmt.Println(string(ob))

  if string(ob) != data {
    panic("Failed assertion")
  }
}
```

and the output
```
[{{ Database path} <nil>} {database [{path path/to/db} {{ Server port} <nil>} {server 9090} {{ Sub comment} <nil>} {sub [{val foo} {{ Slice comment} <nil>}]} {sl [one { Two} two]}]}]
# Database path
database:
  path: path/to/db
  # Server port
  server: 9090
  # Sub comment
  sub:
    val: foo
    # Slice comment
  sl:
  - one
  # Two
  - two

panic: Failed assertion

goroutine 1 [running]:
main.main()
	/Users/buchanae/go/src/scratch/potato/potato.go:37 +0x387
exit status 2
```

Thanks for doing this. I'm hoping to use this to better manage YAML config files. Let me know I can help move this along somehow.

buchanae:
Oh, I actually just noticed that the "Database path" comment is actually bumped up a level, incorrectly, to the "database" key.

Perhaps this is related to the original comment in this PR:

> We don't retain indentation information for comments. Comments are placed at the same indentation level as the previous token.

autrilla:
Yes, that's what's happening here. Keeping the comment where it was
initially place is a harder problem, and it's specially harder considering
iirc currently the library doesn't keep tokens on the same "style" they
were in. So a round trip might change indentation, or even string style
(not particularly sure about this last one).

josephholsten:
@autrilla any progress with this? I'm really keen on replacing my usage of https://pypi.python.org/pypi/yamlfmt/0.1.5 with this

autrilla:
@josephholsten this is probably never going to get to the level of ruaml.yaml's round trip parser. go-yaml doesn't keep style information, and changing it so it does would be hard. Using this to reformat your yaml will remove some information. I'm not sure if this answers your question, but the progress you see here is all the progress we've made.

josephholsten:
ah, that's good to know. how would you feel about a fork that did keep that info? Would it make sense for that to live in a separate package, or do you think that might be worth being part of go-yaml?

autrilla:
Well, this PR has been open for a while now and hasn't received any comments from contributors, so I don't know if there's any plan to ever maintain presentation layer information such as comments. A fork that keeps this info would be great, and you might consider starting from what I worked on. A PR back to mozilla-services' fork that retains more presentation information would be amazing 😄

lox:
There have been a few discussions on comment handling at https://github.com/go-yaml/yaml/issues/132 and https://github.com/go-yaml/yaml/issues/129. 

@niemeyer would the above change be something that might make sense in v3?

niemeyer:
Yes, this is certainly worth looking into after v3 is out.

Preliminary issues are lack of tests and conflicts.

Naatan:
What is the timeline on v3? Very interested in this PR.

khimaros:
I would love to see this PR merged.

niemeyer:
Thank you very much for the PR and for the discussion. It's clear that everybody wants comment decoding and encoding, and historically it's also clear that there's significant interest in preserving some of the document structure to enable mechanical modifications.

As such, I've been integrating comment decoding/encoding into v3, using the intermediate state management that comes with it. The approach used internally is also slightly different from the one here, in the sense comments are not first-class tokens, but rather metadata associated with existing tokens. This makes it much easier to understand, manipulate, and preserve file comments as desired.

Here is a sneak peek of what's coming:

https://twitter.com/gniemeyer/status/1073638853225996288

I still need to handle a few more edge cases, but given the amount of testing so far I'm comfortable that this is going to work.

On that basis, I'm tentatively closing this PR for the time being. Thanks again for the collaboration here.

zzak:
@niemeyer Thank you for the update! I'm very excited to see progress here.

Is there anywhere we can try out an early build of v3?

vdods:
Hi, I'm interested in getting comments in go-yaml as well.  Any word on v3?

Issue #144, opened 2015-11-09
yaml.Marshal panics when serialising [10]bool


davecheney:
yaml.Marshal will panic when fed this value

```
v := struct{ A, B [10]bool } { } 
```

```
PANIC: yaml_test.go:74: yamlSuite.TestWriteYamlWriteGarbage

... Panic: Can't marshal type: [10]bool (PC=0x45B90E)

/home/dfc/go/src/runtime/asm_amd64.s:471
  in call32
/home/dfc/go/src/runtime/panic.go:423
  in gopanic
/home/dfc/src/gopkg.in/yaml.v1/yaml.go:28
  in handleErr
/home/dfc/go/src/runtime/asm_amd64.s:471
  in call32
/home/dfc/go/src/runtime/panic.go:423
  in gopanic
/home/dfc/src/gopkg.in/yaml.v1/encode.go:108
  in encoder.marshal
/home/dfc/src/gopkg.in/yaml.v1/encode.go:141
  in encoder.structv.func1
/home/dfc/src/gopkg.in/yaml.v1/encode.go:155
  in encoder.mappingv
/home/dfc/src/gopkg.in/yaml.v1/encode.go:143
  in encoder.structv
/home/dfc/src/gopkg.in/yaml.v1/encode.go:90
  in encoder.marshal
/home/dfc/src/gopkg.in/yaml.v1/yaml.go:143
  in Marshal
```

Why doesn't yaml.Marshal return this as an error ?

aanm:
Same with `[2]int`

```
panic: cannot marshal type: [2]int [recovered]
        panic: cannot marshal type: [2]int [recovered]
        panic: cannot marshal type: [2]int
```

vinzenz:
This is a duplicate of #130 I can see now

niemeyer:
That's been fixed by https://github.com/go-yaml/yaml/commit/2aba0a492be00f1eb4d95483b08930ebe4968b64.

Issue #130, opened 2015-10-06
Unable to marshal array (not slice)


gotwarlost:
This code panics. (I was trying to marshal a docker container config object that has this construct)

```
package main

import (
    "gopkg.in/yaml.v2"
)

type Config struct {
    ConsoleSize [2]int
}

func main() {
    c := &Config{ConsoleSize: [2]int{10, 20}}
    yaml.Marshal(c)
}
```

niemeyer:
That's been fixed by the commit above.

PR #221, opened 2016-11-02
Implement capability to marshal array (not slice) (Was #130 and #144)

vinzenz:
This should fix the array encoding issue

This closes issues #144 and closes issue #130


niemeyer:
That's been fixed by the commit above.

Issue #214, opened 2016-10-21
Boolean support inconsistent with the YAML 1.2 spec


fabianofranz:
The YAML 1.1 spec [explicitly supported unquoted values like "on/off" or "yes/no" as booleans](http://yaml.org/type/bool.html), so the sample below would translate the third element of `command` to `false`:

```
command:
  - redis-server
  - '--appendonly'
  - no
```

However these boolean representations [were dropped in 1.2 as one of the major changes in the spec](http://www.yaml.org/spec/1.2/spec.html), in favour of just `true` or `false` to represent booleans.

This change is not represented in the go-yaml library, which [still supports all previous forms of booleans](https://github.com/go-yaml/yaml/blob/v2/resolve.go#L36-L41). I understand it's a claim of the project that it supports most of 1.1 and 1.2, however for larger spec changes like this one, it would be important to have the ability to choose if I want to stick to the 1.2 behavior for booleans, and have the `no` element in the example above translated to the string "no".

fabianofranz:
Here's the initial motivation for this in Kubernetes, and one possible idea for "fixing" it:

https://github.com/kubernetes/kubernetes/issues/34146#issuecomment-255821722

jonasfj:
test case
```golang
func TestYAsStringKeyInMap(t *testing.T) {
	var raw map[interface{}]interface{}
	yaml.Unmarshal([]byte(`{x: 10, y: 1}`), &raw)
	if _, ok := raw[true]; ok {
		t.Errorf("read y as yes instead of 'y', result: %#v", raw)
	}
}
```

This is pretty scary stuff...

rogpeppe:
> This is pretty scary stuff...

Welcome to YAML.

niemeyer:
We cannot touch this in v2 as it will break existing code, but we'll make this change on v3.

andig:
@niemeyer I‘ve been bitten by this, too. Could you kindly confirm what the status is with regards to unmarshaling in v3?

**UPDATE** answering myself: https://github.com/go-yaml/yaml/tree/v3

Seems this has been changed and this issue could be closed?

horacimacias:
is this still an issue or can be closed given v3 seems to be already available?

PR #222, opened 2016-11-02
Add support for strict boolean parsing (Was #214)

vinzenz:
This should help with kubernetes/kubernetes#34146

by introducing a new option to restrict the booleans to only `true` and `false`

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -26,6 +26,7 @@ type node struct {
 	implicit     bool
 	children     []*node
 	anchors      map[string]*node
+	data         interface{}
 }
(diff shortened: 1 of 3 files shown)


ghodss:
@vinzenz What is the latest on this PR? Is it up to date with the latest commits? Is it ready to be merged if we can get it reviewed?

niemeyer:
Version 1.2 of the spec constrains booleans to only true and false indeed, and that's the default behavior we'll have in v3, which is going to be out soon.

Issue #157, opened 2016-02-09
go-yaml should handle invalid integers as strings


mbruzek:
Starting a numeric sequence with zero indicates an octal number.  So 01234567 == 342391 (base 10). Numeric sequences that start with zero and contain numbers higher than 7 are invalid integers and should be parsed as strings even when they do not have quotes around them.

``` yaml
key: 01182252
```

01182252 is not a valid octal integer and should be treated as a string. go-yaml attempts to parse the numeric sequence as an integer and this is incorrect.

Several yaml validators and other language yaml packages handle this case correctly such as the Python ruamel package and on line parsers http://yaml-online-parser.appspot.com/

```
package main

import (
    "fmt"
    "gopkg.in/v2/yaml"
)

var a = make(map[string]interface{})

func main() {
    data := []byte(`key: 01234567`)
    err := yaml.Unmarshal(data, a)
    if err != nil {
        fmt.Printf("%v\n", err)
    }
    fmt.Printf("%v\n", a["key"])

    // This value is surrounded by quotes.
    data = []byte(`key: "01182252"`)
    err = yaml.Unmarshal(data, a)
    if err != nil {
        fmt.Printf("%v\n", err)
    }
    fmt.Printf("%v\n", a["key"])

    // This value is not a valid octal integer therefore a string.
    data = []byte(`key: 01182252`)
    err = yaml.Unmarshal(data, a)
    if err != nil {
        fmt.Printf("%v\n", err)
    }
    fmt.Printf("%v\n", a["key"])
}
```

jbiel:
This is an issue for me.

qix:
I'm seeing this issue in an unlucky git commit hash that contained only digits `0787904`.

Python serializes it to an unquoted string, which I don't like _but_ is technically okay I believe.

```
print yaml.safe_dump({'test': '0787904'}, default_flow_style=False)
test: 0787904
```

CpuID:
This would be awesome to have resolved, docker-compose.yml ports are recommended to be defined as strings instead of integers.

dsegan:
FWIW, the spec seems to confirm this is wrong: http://yaml.org/spec/1.2/spec.html#id2803828

Integers are  0 | -? [1-9] [0-9]*

niemeyer:
Merged #171.

howbazaar:
PR #171 doesn't actually fix this issue.

jameinel:
[...]

PR #223, opened 2016-11-03
Fix for issue #157 - go-yaml should handle invalid integers as strings

vinzenz:


niemeyer:
It's not clear to me how this is fixing the referenced issue in a proper way. What's the actual specification of the float format in yaml, and how is this handling it? Where are the tests for it showing details of the problem being fixed?

vinzenz:
> It's not clear to me how this is fixing the referenced issue in a proper way. 

ParseFloat is  behaving oddly by discarding the leading 0 and make the rest of the number a float.

> What's the actual specification of the float format in yaml, and how is this handling it? 

http://yaml.org/spec/1.2/spec.html#tag/repository/float
A leading 0 is only allowed when followed by a dot, which it checks for and then passes the string on to really try a float. (tryFloat)

> Where are the tests for it showing details of the problem being fixed?

I will add some test that would fail before

niemeyer:
Note that this changed from yaml 1.1 to 1.2. It might be okay, but we may also end up breaking existing code. In Python:
```
>>> import yaml
>>> yaml.load("01.1")
1.1
```
Thanks for the tests.

niemeyer:
Curiously, Python float resolution seems broken in other ways:
```
>>> yaml.load("-2E+05")
'-2E+05'
```

vinzenz:
#171  seems to actually fix this already

vinzenz:
And we can also add this as a parser option (Strict floats) like it was introduced as another PR

niemeyer:
Indeed, merged #171. Thanks for your help on this.

PR #240, opened 2017-02-08
Fix dead url for yaml.v2/scannerc.go

xulike666:
I couldn't find the YAML specification on http://yaml.org/spec/cvs/current.html and I update it.


niemeyer:
Thanks!

PR #241, opened 2017-02-15
Update the yamlStyleFloat regex for special case.

### LINKED ISSUE #157: go-yaml should handle invalid integers as strings

howbazaar:
Fixes #157 by updating the regex for float matching.

Instead of allowing mantissa to be any 0-9 sequence, this change means it must be either 0, or not start with a zero.

### DIFF
--- a/resolve.go
+++ b/resolve.go
@@ -81,7 +81,7 @@ func resolvableTag(tag string) bool {
 	return false
 }
 
-var yamlStyleFloat = regexp.MustCompile(`^[-+]?[0-9]*\.?[0-9]+([eE][-+][0-9]+)?$`)
+var yamlStyleFloat = regexp.MustCompile(`^[-+]?(0|[1-9][0-9]*)?(\.[0-9]+)?([eE][-+][0-9]+)?$`)
 
 func resolve(tag string, in string) (rtag string, out interface{}) {
 	if !resolvableTag(tag) {
--- a/decode_test.go
+++ b/decode_test.go
@@ -43,6 +43,9 @@ var unmarshalTests = []struct {
 		"v: 0xA",
 		map[string]interface{}{"v": 10},
 	}, {
+		"v: 01182252",
+		map[string]interface{}{"v": "01182252"},
+	}, {
 		"v: 4294967296",
 		map[string]int64{"v": 4294967296},
 	}, {


wallyworld (approved):

resolve.go
```diff
@@ -81,7 +81,7 @@ func resolvableTag(tag string) bool {
 	return false
 }
 
-var yamlStyleFloat = regexp.MustCompile(`^[-+]?[0-9]*\.?[0-9]+([eE][-+][0-9]+)?$`)
+var yamlStyleFloat = regexp.MustCompile(`^[-+]?(0|[1-9][0-9]*)?(\.[0-9]+)?([eE][-+][0-9]+)?$`)
```
gabriel-samfira, line 84: Shouldn't this be:

```go
// the sign of what follows after "[eE]" should be optional. 
var yamlStyleFloat = regexp.MustCompile(`^[-+]?(0|[1-9][0-9]*)?(\.[0-9]+)?([eE][-+]?[0-9]+)?$`)
```

to allow:

```go
floats := []string{
   "2.2e11",
   "2.2e+11", // same as above, but with an explicit sign
   "2.2e-11",
}
```
?

howbazaar: Well the spec says that all exponents must have explicit sign.

niemeyer:
Per the spec, this isn't right. We approved a different PR and unfortunately that was also not right.

YAML 1.2 actually has this regular expression for floats:

`-? ( 0 | [1-9] [0-9]* ) ( \. [0-9]* )? ( [eE] [-+]? [0-9]+ )?`

We'll adopt that in v3 only to avoid breaking existing code.

PR #242, opened 2017-02-20
Add nowrap tag to ensure that values are outputted on to a single line

icambridge:
This adds a no wrap tag to allow for people to avoid having unwanted line breaks in their values.

It also allows for the adding of other such tags to be done in an easier fashion.


LLParse:
+1 Would like to see this added

niemeyer:
What you want is to change the line width, which is already supported by the backing infrastructure.

But let's please do this in v3.

PR #243, opened 2017-02-28
make TextUnmarshaler errors non-fatal, add line number

fjl:
TextUnmarshaler is commonly used for 'small' types such as IP addresses.
If a parse error occurs with such a type, the current behavior is to
break out of parsing and return that error directly. When package yaml
is used to process configuration files, it is more appropriate to emit a
parser errror with line number info and continue.

This changes API behavior, but I cannot imagine a use case where one
would rely on yaml returning the exact error from UnmarshalText. Support
for TextUnmarshaler isn't even documented.


niemeyer:
We fail on pretty much any incorrect typing, and also when the stock yaml.Unmarshaler fail. It would be awkward to not error when the data being provided is completely broken.

Note that you can do that in your code by implementing a custom unmarshaler, though.

Issue #245, opened 2017-03-02
YML document "<<: value" cannot be parsed


jochenehret:
Hi,

we want to parse yml documents which contain "<<" as map key names. These are used in the context of [spiff yml file merging](https://github.com/cloudfoundry-incubator/spiff#--merge-). If we try to parse the document:
```
<<: value
```
we get the error `yaml: map merge requires map or sequence of maps as the value`.
To our understanding, the yml document is correct (key contains printable character). Other tools can parse the document without problems:
```
$ echo "<<: value" | yaml2json | jq .
{
  "<<": "value"
}
```
Can you please check and fix?

Thanks and Best Regards,

  Jochen.

asheidan:
According to the YAML cookbook I think you need to use "<<" as key instead

```yaml
"<<": value
```

http://yaml.org/YAML_for_ruby.html#merge_key

rogpeppe:
The current behaviour seems reasonable to me, but I'm open to arguments the other way. Can anyone cite the relevant part of the spec that would allow an unquoted `<<` as a mapping key?

rogpeppe:
After looking into a bit more, I agree with @asheidan. The actual bug here is that << does not get quoted when marshaled as a mapping key. We should probably make YAML quote "<<" when used as a mapping key, and export a MergeKey type or value so that external code can deliberately elect to print an unquoted << value (they already have the option to use the merge tag explicitly, however, although that will look awkward).

lox:
Given https://github.com/go-yaml/yaml/pull/313 is merged, can this be closed now?

jochenehret:
I've tried to parse
```
"<<": value
```
with the fix https://github.com/go-yaml/yaml/pull/313/commits/0e9d51246b84485169bff330f09d6d8f55d819cc, but I still get the error `"yaml: map merge requires map or sequence of maps as the value"`.

However, this issue is not really critical for us, so it would be ok if you close it.

Best Regards,

  Jochen.

ddollar:
I just ran into what I think is this issue when trying to use a merge key.

This parses as expected and merges the maps:

```
foo: &foo
  bar: bar

baz:
  "<<": *foo
  qux: quux
```

but this does not:

```
foo: &foo
  bar: bar

baz:
  <<: *foo
  qux: quux
```

From what I can tell the spec does show unquoted merge keys should work:

http://yaml.org/type/merge.html

ddollar:
[...]

PR #247, opened 2017-03-14
Fix for issue #245

philippthun:
As outlined in issue #245 we have YAML files containing "<<" as key names for values that should (and cannot be merged) according to http://yaml.org/type/merge.html. The proposed fix checks if the value node is of type 'scalar node' and then does not try to merge; unless explicitly requested (i.e. by adding the !!merge tag).

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -546,7 +546,7 @@ func (d *decoder) mapping(n *node, out reflect.Value) (good bool) {
 	}
 	l := len(n.children)
 	for i := 0; i < l; i += 2 {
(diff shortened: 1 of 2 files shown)


rogpeppe:
As I commented in the issue, I'm not sure this quite the right fix.

lox:
Given https://github.com/go-yaml/yaml/pull/313 is merged, can this be closed now?

niemeyer:
This doesn't look right. You need to quote it if you want it as a string.

Note that the test proposed doesn't pass in Python either.

Issue #154, opened 2015-12-28
YAML spec violation: Duplicate keys are silently ignored, and parser selects incorrect key


RichiH:
As per https://github.com/prometheus/prometheus/issues/1275 :

If you run with a configuration like:

```
scrape_configs: <<<<<<<<<<<<<< this is discarded
  - job_name: 'prometheus_system'
    target_groups:
    - targets: ['localhost:9100']

scrape_configs:
  - job_name: 'foo_system'
    target_groups:
    - targets: ['foot:9100']
```

the former `scrape_configs` is silently discarded. While that's the obvious case, constructs like

```
scrape_configs: <<<<<<<<<<<<<< this is discarded
  - job_name: 'prometheus_system'
    target_groups:
    - targets: ['localhost:9100']

rule_files:
  - 'prometheus.rules'

scrape_configs:
  - job_name: 'foo_system'
    target_groups:
    - targets: ['foo:9100']
```

are silently discarded as well.

Raising an error would be appreciated, and suggested by the YAML specs, in this situation.

Also, as @fabxc notes, http://yaml.org/spec/1.1/#id932806 says that the first, not the last, occurence should be valid.

davidje13:
We have created a new pull request to resolve this; it includes tests for the situation described, and tests for valid cases (e.g. duplicate values are OK; only check for duplicate keys)

stub42:
Can this land on the v2 branch, or does this fix break backwards compatibility and need to land on a v3 branch?

jsha:
Any update on possibly merging to the v2 branch?

4a6f656c:
I believe this is addressed in the v2 branch via [UnmarshalStrict](http://godoc.org/gopkg.in/yaml.v2/#UnmarshalStrict). Although be aware that this also requires that there are fields present in the struct for all fields in the YAML data.

@rogpeppe - this should probably have been tagged as fixed in https://github.com/go-yaml/yaml/pull/307 - unless there are plans to add duplicate key detection to non-strict mode, then I suspect this should be closed.

niemeyer:
I will review this behavior and perhaps make it standard in v3. But yet, today that new strict mode already handles the case discussed here.

davidje13:
@niemeyer in strict mode yes this sounds resolved, but if the non-strict mode doesn't error, I'd say it should only be considered resolved once it follows the spec (as mentioned in the original issue)

[...]

PR #248, opened 2017-03-24
Reject duplicate keys (addresses #154)

davidje13:
This addresses https://github.com/go-yaml/yaml/issues/154 (reject duplicate keys in mappings as per [the spec](http://yaml.org/spec/1.2/spec.html#id2764044))

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -176,8 +176,16 @@ func (p *parser) mapping() *node {
 	n := p.node(mappingNode)
 	p.anchor(n, p.event.anchor)
 	p.skip()
+	duplicates := make(map[string]bool)
 	for p.event.typ != yaml_MAPPING_END_EVENT {
-		n.children = append(n.children, p.parse(), p.parse())
+		key := p.parse()
+		value := p.parse()
+		if _, exists := duplicates[key.value]; exists {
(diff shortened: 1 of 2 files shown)


jsha:
@davidje13 would you merge master? There's a conflict in decode_test.go.

@rogpeppe @gbbr I'd love to see this merged, since we have a need to reject duplicate keys in config files for Let's Encrypt's Boulder project. Do you think you might have time to take a look soon? Thanks!

davidje13:
@jsha resolved (wasn't much of a conflict; just that each branch had added a distinct test in the same location), though it seems to be failing the build now due to an unrelated issue (`./example_embedded_test.go:34: Fatal call has possible formatting directive %v`); I guess the checker updated its rules to be more strict within the last 23 days (since the last successful build).

I don't have much faith that this will get merged though; there doesn't seem to be any interest in this bug from the maintainers. The [issue](https://github.com/go-yaml/yaml/issues/154) has been open since December 2015 (2 years!) and there have been 2 viable PRs sitting for about a year without any comment from the maintainers, despite multiple downstream projects complaining about it.

rogpeppe:
Superceded by #307

PR #252, opened 2017-04-24
Add support for an octal type

xyproto:
File permissions in YAML files are more intuitive when written like `0660` rather than `432`.

I see you already have a check in place for `time.Duration`, so I added one for an `Octal` type too.

This is only for writing numbers as octal, `go-yaml/yaml` can already read octal numbers just fine.

I'm not saying this is the perfect solution, but it worked for me.

Cheers!


rogpeppe:
I'm not sure that a special case for time.Duration should exist. It's straightforward to write your own type that implements encoding.TextMarshaler and TextUnmarshaler by reading and writing octal values, so I'd suggest that approach.

PR #253, opened 2017-04-27
docs: note need for public struct fields

heldtogether:
It took me longer than it should have to realise that struct fields need to be public for unmarshal to work correctly. Just adding a quick note to that effect.


rogpeppe (approved):
Worth doing, thanks.

PR #256, opened 2017-05-06
encoder.marshal panics on nil pointers

sb-akd:
func (e *encoder) marshal(tag string, in reflect.Value) should not try to perform any operations (call MarshalYAML() or MarshalText()) on nil pointers


doroginin:
+1

rogpeppe (approved):
LGTM, thanks (I'll add a test in another PR).

PR #260, opened 2017-05-10
Marshal fixed-length arrays correctly

### LINKED ISSUE #259: Marshaling a [2]uint results in a panic
I'm currently encountering the following panic:

```
panic: cannot marshal type: [2]uint [recovered]
        panic: cannot marshal type: [2]uint

goroutine 1 [running]:
github.com/go-yaml/yaml.handleErr(0xc420524818)
        /home/paralin/Documents/goworkspace/src/github.com/go-yaml/yaml/yaml.go:153 +0x95
```

Is it not reasonable that YAML should be able to handle unsigned integers?...

paralin:
This fixes #259

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -99,6 +99,8 @@ func (e *encoder) marshal(tag string, in reflect.Value) {
 		}
 	case reflect.Struct:
 		e.structv(tag, in)
+	case reflect.Array:
+		fallthrough
 	case reflect.Slice:
 		if in.Type().Elem() == mapItemType {
 			e.itemsv(tag, in)


rogpeppe (changes requested):
Thanks for doing this, but please add some tests for this.

Also, please check what happens when unmarshaling - I
think it should be possible to unmarshal into a fixed length array
so we can unmarshal into the thing we've marshaled.

Perhaps check what encoding/json does for guidance as to appropriate
semantics in that case.

niemeyer:
That's been fixed by the commit above.

PR #210, opened 2016-10-10
Modified: Don't ignore unspecified fields (from #136)

gemcdaniel:
This is a slightly modified version of @douglaswth PR [#146](https://github.com/go-yaml/yaml/pull/146) which solves @oec's [#136](https://github.com/go-yaml/yaml/issues/136) issue.

PR #262, opened 2017-06-13
Add UnmarshalStrict method returning error if yaml has fields that do not exist in structure

wupeka:
Currently, there's no way of knowing if the whole YAML has been parsed, new UnmarshalStrict method returns an error if there are fields in YAML that were missing from the structure.

### DIFF
--- a/yaml.go
+++ b/yaml.go
@@ -77,8 +77,19 @@ type Marshaler interface {
 // supported tag options.
 //
 func Unmarshal(in []byte, out interface{}) (err error) {
+	return unmarshal(in, out, false)
+}
+
+// UnmarshalStrict is like Unmarshal except that any fields that are found
+// in the data that do not have corresponding struct members will result in
+// an error.
+func UnmarshalStrict(in []byte, out interface{}) (err error) {
+	return unmarshal(in, out, true)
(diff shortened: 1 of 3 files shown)


yaml.go
```diff
@@ -77,8 +77,19 @@ type Marshaler interface {
 // supported tag options.
 //
 func Unmarshal(in []byte, out interface{}) (err error) {
+	return unmarshal(in, out, false)
+}
+
+// UnmarshalStrict decodes the provided yaml just like Unmarshal but returns
+// type errors also for fields that exists in YAML but are missing from provided
```
rogpeppe: // UnmarshalStrict is like Unmarshal except that any fields that are found
// in the data that do not have corresponding struct members will result in an error.

?

rogpeppe (approved):
LGTM with one suggestion, and the thought that it might be nicer if the error could show *all* the extra fields rather than just one. But that's probably a future improvement.

douglaswth:
Does this fix #136/#146? I noticed that #210 was closed probably because of this PR.

jk2K:
has fixed https://github.com/go-yaml/yaml/pull/146

PR #268, opened 2017-07-17
Expose parse

firelizzard18:
I need to parse YAML documents without unmarshalling into a structure.


niemeyer:
The parser is really an implementation detail which shouldn't be exposed at this point. If you need to reuse the code, I suggest just copying it over.

firelizzard18:
So what you're saying is that this package is not "YAML support for the Go language" but rather just encoding/decoding support with absolutely no other YAML support?

PR #269, opened 2017-07-17
Add support for marshaling regexp.

jfrabaute:
As it is done for time.Duration, we have structs with fields of type regexp.Regexp so it could be interesting to unmarshal those ones directly using the yaml parser.

This is a pull request adding this feature.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math"
 	"reflect"
+	"regexp"
 	"strconv"
 	"time"
 )
@@ -198,6 +199,7 @@ var (
 	durationType   = reflect.TypeOf(time.Duration(0))
 	defaultMapType = reflect.TypeOf(map[interface{}]interface{}{})
 	ifaceType      = defaultMapType.Elem()
+	regexpType     = reflect.TypeOf(regexp.Regexp{})
 )
 
 func newDecoder(strict bool) *decoder {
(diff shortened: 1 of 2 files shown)


jfrabaute:
Any chance to have this merged?

Thanks.

rogpeppe:
I'm not sure that this is above the bar for inclusion. There are many possible types that could be given special treatment, and I'm not sure that regexp is one of them. To be honest, I wouldn't include Duration itself unless it was already there - it's not actually documented.

You can always do something like this yourself by defining a custom type which implements yaml.Marshaler or encoding.TextMarshaler and using that in the types you're marshaling.

jfrabaute:
> You can always do something like this yourself by defining a custom type which implements yaml.Marshaler or encoding.TextMarshaler and using that in the types you're marshaling.

Yes, that's what I did. I have to have a type which is a regexp but just a custom one to have be able to implement yaml.Marshaler. Not perfect but it works.

I understand your point. I'll keep it like that with the custom type.

Thanks for taking the time to answer.

rwxrob:
Perhaps it is time to reopen this topic since the core `encoding/json` marshaling handles `*regexp.Regexp` from YAML strings directly. This is no longer an edge case, it's a core piece of JSON marshaling functionality that is absent from yaml.v3 making it incompatible with the standard package, in fact, we just ran into this discrepency in a production development project and were forced to add a MarshalYAML where a similar one would not have been needed for JSON instead.

PR #272, opened 2017-07-21
Make tag scanning code slightly cleaner.

rogpeppe:
The current code uses "length" as a proxy for "has a tag"
but the length doesn't actually match the length of the
tag when there's a %-encoded character, so use a boolean
instead.

Also fix a panic found when trying to write a test for the change.
Aside: seems that %-encoded tag handles are not allowed by
the YAML spec.


mattyw:
LGTM

Issue #284, opened 2017-11-03
`UnmarshalStrict()` doesn't error on duplicate keys


aeijdenberg:
I spent yesterday afternoon tracking down a bug in our Concourse configuration whereby the root cause was a YAML file containing:

```yaml
resource_types:
- name: foo
- name: bar

... hundreds of unrelated lines later ...

resource_types:
- name: baz
```

where the final `resource_types` effectively deleted the first.

I thought it would be within the spirit of what `UnmarshalStrict()` sounds like it should do, if it would return an error if a key is attempted to be set in a map (or struct) that already has value.

PR forthcoming for your consideration...

aeijdenberg:
Related: <https://github.com/go-yaml/yaml/issues/154>

xunwu:
I am facing the same issue here, I think it should throw an error if duplicate keys are found, rather than silently replacing the value

rogpeppe:
Fixed by #307.

PR #285, opened 2017-11-03
Return an error in strict mode if yaml contains duplicate values for a map or struct

aeijdenberg:
Return an error in strict mode if yaml contains duplicate values for a map or struct.

Fixes <https://github.com/go-yaml/yaml/issues/284>.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -538,6 +538,16 @@ func (d *decoder) sequence(n *node, out reflect.Value) (good bool) {
 	return true
 }
 
+func (d *decoder) setMapIndex(n *node, out, k, v reflect.Value) {
+	if d.strict {
+		if out.MapIndex(k) != zeroValue {
+			d.terrors = append(d.terrors, fmt.Sprintf("line %d: field %s already set in map", n.line+1, k.String()))
+			return
+		}
+	}
+	out.SetMapIndex(k, v)
+}
+
 func (d *decoder) mapping(n *node, out reflect.Value) (good bool) {
 	switch out.Kind() {
 	case reflect.Struct:
@@ -592,7 +602,7 @@ func (d *decoder) mapping(n *node, out reflect.Value) (good bool) {
 			}
 			e := reflect.New(et).Elem()
 			if d.unmarshal(n.children[i+1], e) {
-				out.SetMapIndex(k, e)
+				d.setMapIndex(n, out, k, e)
 			}
 		}
 	}
@@ -647,6 +657,7 @@ func (d *decoder) mappingStruct(n *node, out reflect.Value) (good bool) {
 		elemType = inlineMap.Type().Elem()
 	}
 
+	doneFieldNames := make(map[string]bool)
 	for i := 0; i < l; i += 2 {
 		ni := n.children[i]
 		if isMerge(ni) {
@@ -656,7 +667,15 @@ func (d *decoder) mappingStruct(n *node, out reflect.Value) (good bool) {
 		if !d.unmarshal(ni, name) {
 			continue
 		}
-		if info, ok := sinfo.FieldsMap[name.String()]; ok {
+		ns := name.String()
+		if d.strict {
+			if doneFieldNames[ns] {
+				d.terrors = append(d.terrors, fmt.Sprintf("line %d: field %s already set in type %s", n.line+1, ns, out.Type()))
+				break
+			}
+			doneFieldNames[ns] = true
+		}
+		if info, ok := sinfo.FieldsMap[ns]; ok {
 			var field reflect.Value
 			if info.Inline == nil {
 				field = out.Field(info.Num)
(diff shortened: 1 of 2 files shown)


decode.go
```diff
@@ -625,7 +635,17 @@ func (d *decoder) mappingStruct(n *node, out reflect.Value) (good bool) {
 		if !d.unmarshal(ni, name) {
 			continue
 		}
-		if info, ok := sinfo.FieldsMap[name.String()]; ok {
+		ns := name.String()
+		if d.strict {
+			for _, df := range doneFields {
```
rogpeppe: This linear search seems a little more heavyweight than it needs to be. If we keep a unique field index in each fieldInfo (the index into fieldsList could be used when building up the struct info in getStructInfo, perhaps), then doneFields could be a []bool, and checking whether a field has already been done would just require indexing into the slice.

decode.go
```diff
@@ -625,7 +635,17 @@ func (d *decoder) mappingStruct(n *node, out reflect.Value) (good bool) {
...
+				if ns == df {
+					d.terrors = append(d.terrors, fmt.Sprintf("line %d: field %s already set in struct %s", n.line+1, ns, out.Type()))
```
rogpeppe: perhaps `set in type %s` so that we don't stutter when the type is an unnamed struct?

decode_test.go
```diff
@@ -977,14 +978,21 @@ func (s *S) TestUnmarshalSliceOnPreset(c *C) {
...
 	c.Check(err, IsNil)
 	err = yaml.Unmarshal([]byte("a: 1\nb: 2\nc: 3"), &v)
 	c.Check(err, IsNil)
 	err = yaml.UnmarshalStrict([]byte("a: 1\nb: 2\nc: 3"), &v)
-	c.Check(err, ErrorMatches, "yaml: unmarshal errors:\n  line 1: field c not found in struct struct { A int; B int }")
+	c.Check(err, ErrorMatches, "yaml: unmarshal errors:\n  line 1: field c not found in struct struct { A int; B int; Z map\\[string\\]string }")
+	err = yaml.UnmarshalStrict([]byte("a: 1\na: 2"), &v)
+	c.Check(err, ErrorMatches, "yaml: unmarshal errors:\n  line 1: field a already set in struct struct { A int; B int; Z map\\[string\\]string }")
+	err = yaml.UnmarshalStrict([]byte("z:\n  a: 1\n  a: 1"), &v)
+	c.Check(err, ErrorMatches, "yaml: unmarshal errors:\n  line 2: field a already set in map")
```
rogpeppe: Let's have a test for unmarshaling duplicate values into a struct with inline members too, please (including an inline struct and a map)

decode.go
```diff
@@ -507,6 +507,15 @@ func (d *decoder) sequence(n *node, out reflect.Value) (good bool) {
 	return true
 }
 
+func (d *decoder) setMapIndex(n *node, out, k, v reflect.Value) {
+	if d.strict {
+		if out.MapIndex(k) != zeroValue {
+			d.terrors = append(d.terrors, fmt.Sprintf("line %d: field %s already set in map", n.line+1, k.String()))
```
rogpeppe, line 544: Shouldn't this return without setting the value? (not that it matters that much I guess)

rogpeppe (changes requested):
Thanks very much for this. This looks reasonable, but I'd prefer to use a lighter weight solution for keeping track of already used struct fields. Also, this needs an update to the UnmarshalStrict docs to document the new behaviour.

decode.go
aeijdenberg: While it probably doesn't matter much, I agree it shouldn't set a value after it's decided that's an error. Fixing as suggested.

decode.go
aeijdenberg: I nearly followed this, but got a bit lost trying to understand the `InlineMap` stuff...

Instead I changed `doneFields` to a `map[string]bool` - which I think at minimum makes the code easier to read.

decode.go
aeijdenberg: Fixed.

decode_test.go
aeijdenberg: Added, I think - I hadn't seen this form before so not 100% confident.

jsha:
@rogpeppe do you have time for a re-review? On the Let's Encrypt Boulder project we'd very much like to enable detection of duplicate keys in YAML maps. Thanks!

jsha:
@aeijdenberg it looks like this has conflicts now; can you merge master and resolve the conflicts?

aeijdenberg:
@jsha - I've just rebased and squashed on the latest `devel` (I think that's the active development branch).

rogpeppe:
Superceded by PR #307. Thanks very much for this anyway - it pointed me in exactly the right direction.

Issue #265, opened 2017-06-28
~ does not decode as nil for pointer types


ghost:
For example, `Value1` is decoded as an empty string when "~" is used instead of "null":

```go
type T1 struct {
	Value1 *string
}

func main() {
	testinfo := map[string]string{
		"canonical": `value1: ~`,
		"english":   `value1: null`,
	}
	for testname, data := range testinfo {
		var foo T1
		if err := yaml.Unmarshal([]byte(data), &foo); err != nil {
			panic(err)
		}
		fmt.Println("==", testname, "==")
		fmt.Println("input: ", data)
		fmt.Printf("output: %+v\n\n", foo)
	}
}
```

A simple fix [decode.go](https://github.com/go-yaml/yaml/blob/v2/decode.go#L253) works:

```go
	if n.tag == yaml_NULL_TAG || n.kind == scalarNode && n.tag == "" && (n.value == "null" || n.value == "~" || n.value == "" && n.implicit) {
```

All current tests pass, including the one above. I believe #109 is related as this code fixes it as well.  I guess someone forgot about fixing this (or it wasn't a priority, which is understandable.)

fschuetz04:
It would be best to open a pull request for that, maybe the maintainers will merge it.

rogpeppe:
Closed by #289

PR #289, opened 2017-11-12
Support usage of "~" to indicate null values for pointer types

### LINKED ISSUE #109: ~ Null handled incorrectly for *string
When decoding into a *string, null is handled correctly, but ~ is not.

To confirm this, I added the following two tests to `unmarshalTests` in decode_test.go:

``` go
{
    "canonical: ~",
    map[string]*string{"canonical": nil},
}, {
    "english: null",
    map[string]*string{"english": nil},
},
```

The first test fails, but the second test passes. My understanding was that `~` and `null` are semantically equivalent in yaml.

rliebz:
This is an implementation of the fix described in #265, and resolves #109 as well.


rogpeppe:
LGTM; thanks very much.

PR #274, opened 2017-07-22
Rename LICENSE to NOTICE & Add Apache Text in LICENSE

king-jam:
This resolves https://github.com/go-yaml/yaml/issues/235

PR #292, opened 2017-12-12
Add NOTICE back in to resolve loss of Canonical entry.

king-jam:
Resolves: https://github.com/go-yaml/yaml/pull/274#issuecomment-351200873


rogpeppe:
Thanks for this.

PR #288, opened 2017-11-12
Stream Encoder/Decoder like json/xml

im-kulikov:
* workaround implementation of Encoder/Decoder
* merging with master branch
* typos
* formatting
* fixes for linter-warnings
* fixes for tests

PR #306, opened 2018-01-06
implement Encoder

rogpeppe:
This implements a stream encoding type that can encode multiple documents, following the same API as encoding/json.Encoder.

One potentially controversial change is the removal of this special case in encoder.emit:

    if !yaml_emitter_emit(&e.emitter, &e.event) && e.event.typ != yaml_DOCUMENT_END_EVENT && e.event.typ != yaml_STREAM_END_EVENT {

It's not clear why this special case existed, as it causes the error when writing the end of a document (and some of the previous tokens) to be ignored. Hopefully this will not cause any client code to start breaking... @niemeyer Perhaps you might have some background here?

We also remove some of the UTF-16 writer logic, because the output encoding is always set to UTF-8.

### DIFF
--- a/apic.go
+++ b/apic.go
@@ -80,14 +80,13 @@ func yaml_parser_set_encoding(parser *yaml_parser_t, encoding yaml_encoding_t) {
 }
 
 // Create a new emitter object.
-func yaml_emitter_initialize(emitter *yaml_emitter_t) bool {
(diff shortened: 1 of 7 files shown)


rogpeppe:
@im-kulikov would you care to review this PR? It's an alternative to #288.

PR #307, opened 2018-01-07
Enforce key uniqueness in strict mode

### LINKED ISSUE #284: `UnmarshalStrict()` doesn't error on duplicate keys

rogpeppe:
This is similar to PR #285 but somewhat more efficient
because it does not require a new map to be created
for every struct unmarshal.

Fixes #284.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,9 @@ go:
     - 1.4
     - 1.5
     - 1.6
+    - 1.7
+    - 1.8
+    - 1.9
     - tip
 
 go_import_path: gopkg.in/yaml.v2
--- a/decode.go
+++ b/decode.go
@@ -580,12 +580,20 @@ func (d *decoder) mapping(n *node, out reflect.Value) (good bool) {
 			}
 			e := reflect.New(et).Elem()
 			if d.unmarshal(n.children[i+1], e) {
-				out.SetMapIndex(k, e)
+				d.setMapIndex(n.children[i+1], out, k, e)
 			}
 		}
 	}
 	d.mapType = mapType
 	return true
+}
+
+func (d *decoder) setMapIndex(n *node, out, k, v reflect.Value) {
+	if d.strict && out.MapIndex(k) != zeroValue {
+		d.terrors = append(d.terrors, fmt.Sprintf("line %d: key %#v already set in map", n.line+1, k.Interface()))
+		return
+	}
+	out.SetMapIndex(k, v)
 }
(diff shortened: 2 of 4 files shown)


rogpeppe:
@aeijdenberg @jsha Wanna take a look?

aeijdenberg:
Looks good to me, @rogpeppe

aeijdenberg:
@rogpeppe , we'd like to use this in Concourse (https://github.com/concourse/fly/pull/201), but are currently using `gopkg.in/yaml.v2`. Are you likely to cherrypick this to `yaml.v2`?

(it isn't clear to me which bug fixes get merged and which don't - in this case, I don't think it changes the API surface...)

Thanks!

rogpeppe:
My plan was to merge devel into v2 soon, although I had a few more changes planned first. If you're using a dependency manager such as dep, you can always use the devel branch anyway (locking dependencies is always a good idea :-]).

gonix:
This change breaks backwards compatibility in `gopkg.in/yaml.v2`. Our use case is to have default values predefined in go and override them in YAML.

Example:
```
package main

import (
	"fmt"

	yaml "gopkg.in/yaml.v2"
)

type Config struct {
	Number     uint
	Properties map[string]string
}

func main() {
	config := Config{
		Number: 1,
		Properties: map[string]string{
			"foo": "bar",
		},
	}
	userYaml := `
    number: 123
    properties:
        foo: potato
        banana: bar
    `
	err := yaml.UnmarshalStrict([]byte(userYaml), &config)
	if err != nil {
		fmt.Println("ERROR:", err)
	} else {
		fmt.Println(config.Properties)
	}
}
```

Result before this change (gopkg.in/yaml.v2 @ 9e27074):
```
$ go run example.go
map[foo:potato banana:bar]
```

Result after this change (gopkg.in/yaml.v2 @ 77b384e):
```
$ go run example.go
ERROR: yaml: unmarshal errors:
  line 4: key "foo" already set in map
```

PR #309, opened 2018-01-08
Add support for alternative timestamp formats

rogpeppe:
[This is a duplicate of #273 but applied against the devel branch]

yaml.org describes several alternative formats for timestamps not
covered by the default implementation of time.UnmarshalText(). A subset
of these is covered by this PR - future work can enable full
support for all formats.

Timestamps will only be collected if there is an explicit timestamp
tag (ie, not !!str) or if implicit type detection is enabled.

This PR builds on @anthonybishopric's foundations
(see PR #117).

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -484,6 +484,11 @@ func (d *decoder) scalar(n *node, out reflect.Value) (good bool) {
 			out.SetFloat(resolved)
 			good = true
 		}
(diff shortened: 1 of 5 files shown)


rogpeppe:
This PR was landed accidentally, and it breaks code, so devel has been rewound to remove the commit and this will be re-proposed.

PR #313, opened 2018-01-09
quote "<<" keys on output

rogpeppe:
This goes some way to fixing issue #245 - at least
the round trip behaviour will work OK now,
although it's still not possible to read externally-created
YAML which uses << as a map key.

Technically this could be considered an API-breaking change,
because existing code might be relying on the fact that << is not
quoted on output. To enable any such users to work around the
new behaviour (which I'm pretty sure is correct), we introduce a new
MergeTag value which can be used to produce an unquoted merge tag
value.


mhilton (approved):
LGTM

PR #314, opened 2018-01-09
use 64 bit precision when formatting float values

rogpeppe:
It seems that the use of 32 bits here was an oversight.


prashantv:
@rogpeppe  This caused an issue when formatting float32 values (see #352), I put up a fix in #353, would appreciate a review, thanks!

Issue #244, opened 2017-03-02
omitempty flag skipping valid time.Time 


bbengfort:
I'm having a slight problem that I'd be happy to investigate and submit a PR for, but wanted to get your feedback first. It looks like time.Time values inside of a struct are not being serialized with the `omitempty` flag, even though they contain a valid time. My test case:

```go
type Record struct {
    Name string `yaml:"name,omitempty"` 
    Created time.Time `yaml:"created,omitempty"` 
}

func main() {
    records := []*Record{&Record{"foo", time.Now()}, &Record{"bar", time.Now()}}
    data, _ := yaml.Marshal(records)
    ioutil.WriteFile("records.yaml", data, 0644) 
}
```

The output for this is:

```yaml
- name: foo
- name: bar
```

Whereas the output for a `Record` struct defined without `omitempty` is:

```yaml
- name: foo
  created: 2017-03-02T07:58:08.116971948-05:00
- name: bar
  created: 2017-03-02T07:58:08.116972021-05:00
```

The desired behavior here is that if the `Created` field is the zero value for time, namely `time.Time{}`, then the created value isn't serialized, otherwise it is. In this case, time.Now() adds a valid time to the struct.

bbengfort:
Some more information, it looks like the source of the error is the [`isZero()`](https://github.com/go-yaml/yaml/blob/v2/yaml.go#L315) function, though I have not tested this. The time module has a non-standard zero value (e.g. not nil), and as a result provides it's own [`IsZero`](https://golang.org/pkg/time/#Time.IsZero) method. When the yaml method sees the time.Time struct, it iterates through all of its fields to check if they're zero, but since all the time fields are private, it declares the struct zeroed. 

I did some brief research into the reflection package; it looks like we'd have to modify the struct case to detect if it was a time.Time, then use the time.Time.IsZero method in order to have this work correctly; but at the moment I'm not sure how to do that.

rogpeppe:
Fixed by PR #315

bbengfort:
Cheers!

PR #315, opened 2018-01-09
omitempty use IsZero

rogpeppe:
As per issue #244, time.Time does not work well with
omitempty, because it has no public fields.

We make the omitempty logic consult the IsZero
method, if available, to determine whether to omit
a value.

Omitting objects with private fields seems a bit
dubious, but since that's existing behaviour,
we'll keep it.

Fixes issue #244.


mhilton (approved):
LGTM I have some vague recollection that a similar change was rolled back in the go encoding/json package because it had some backwards compatibility issue (I'm probably misremembering).

PR #316, opened 2018-01-09
Decode into structs provided at runtime

IngmarStein:
The current version creates new maps when asked to decode into `interface{}` values. Example:

```
package main

import (
	"fmt"

	"gopkg.in/yaml.v2"
)

type Parent struct {
	Child interface{} `yaml:"child"`
}

type ChildStruct struct {
	Foo int `yaml:"foo"`
}

func main() {
	var data = `
---
child:
  foo: 42
`

	out := Parent{Child: &ChildStruct{}}
	yaml.Unmarshal([]byte(data), &out)

	fmt.Printf("Out: %+v\n", out)
	fmt.Printf("Child: type: %T, value: %+v\n", out.Child, out.Child)

	// expected:
	// Out: {Child:0xc4200142f0}
	// Child: type: *main.ChildStruct, value: &{Foo:42}

	// current:
	// Out: {Child:map[foo:42]}
	// Child: type: map[interface {}]interface {}, value: map[foo:42]
}
```

This PR changes the decoder to use the provided struct pointed to by the interface. This matches the behavior of `encoding/json`.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -711,6 +711,9 @@ func (d *decoder) mappingStruct(n *node, out reflect.Value) (good bool) {
 			} else {
 				field = out.FieldByIndex(info.Inline)
 			}
+			if field.Kind() == reflect.Interface && field.Elem().Kind() == reflect.Ptr {
+				field = field.Elem()
+			}
 			d.unmarshal(n.children[i+1], field)
 		} else if sinfo.InlineMap != -1 {
 			if inlineMap.IsNil() {
--- a/decode_test.go
+++ b/decode_test.go
@@ -1242,6 +1242,26 @@ func (t *textUnmarshaler) UnmarshalText(s []byte) error {
 	return nil
 }
 
+func (s *S) TestDynamic(c *C) {
+	type Parent struct {
+		Child interface{} `yaml:"child"`
+	}
+
+	type ChildStruct struct {
+		Foo int `yaml:"foo"`
+	}
+
+	data := []byte("child:\n  foo: 42")
+
+	out := Parent{Child: &ChildStruct{}}
+	yaml.Unmarshal([]byte(data), &out)
+
+	c.Assert(out, DeepEquals, Parent{Child: &ChildStruct{Foo: 42}})
+	if _, ok := out.Child.(*ChildStruct); !ok {
+		c.Fail()
+	}
+}
+
 //var data []byte
 //func init() {
 //	var err error


rogpeppe:
I don't know any other unmarshaling package that works like you're proposing. Can you think of one? What's the specific use case for this? Usually we're unmarshaling into an empty value.

IngmarStein:
`encoding/json` works like this.

My use case is: I have a struct that contains common configuration options used by multiple services. Each service has another struct for its own settings which it can then just plug into a field of the common struct. Something similar to:

```
// common
struct Config {
  Setting1 string
  Setting2 string
  CustomSettings: interface{}
}

// in service1:
struct Service1Config {
  Custom1 int
  Custom2 bool
}

service1Config := Config{CustomSettings: &Service1Config{}}
yaml.Unmarshal(data, &service1Config)
```

IngmarStein:
Here's a playground with the original example using encoding/json: https://play.golang.org/p/_MffF-A8rTh

IngmarStein:
What's your take on this, @rogpeppe? Does `go-yaml` generally aim to behave similarly to `encoding/json` or is it not supposed to be a drop-in replacement?

rogpeppe:
After some thought and experimentation, I'm pretty sure that this isn't a good idea.

Firstly, you can actually do what you want already: https://play.golang.org/p/0Z-HNopJ39l

But if you start to play with that example a bit, it doesn't work very well.
For example, what if you want to have an slice of interfaces to unmarshal into? Even if you provide a slice with nicely filled-out interfaces, the unmarshaler will create a new slice, ignoring the values in it, and I think that's quite reasonable.

This issue is actually a classic example of something that in encoding/json is solved neatly by json.RawMessage. Unfortunately json.RawMessage isn't quite as easy to do in YAML (although I have some ideas), but I'm fairly sure there are packages around that will unmarshal arbitrary Go types into other struct types, so you could unmarshal into interface{} and then use one of those.

IngmarStein:
`encoding/json` unmarshals into a slice of interfaces just fine: https://play.golang.org/p/Om-B-gkHd00
I think the underlying difference is that go-yaml relies on the static type information whereas `encoding/json` uses the actual value to recurse into nested types.

rogpeppe:
It's interesting that encoding/json works that way - I hadn't realised that. It's not that consistent though - for example it doesn't work with maps: https://play.golang.org/p/0QG3QH4K_sP

I'll continue to think about this.

IngmarStein:
Yeah, I think there might be a more profound change needed to replicate `enconding/json`'s behavior. This PR only enables my specific use case but a more generic solution would be preferable.

niemeyer:
That's incompatible with the current behavior, so we cannot merge in v2. 

Will consider the behavior for v3.

Issue #325, opened 2018-02-03
Inconsistency around "<<:" usage between map and struct


dtbartle:
```
package main
import (
    "fmt"
    "gopkg.in/yaml.v2"
)
var x = `
---
common: &common
  dbs:
    db1: abc
foo:
  <<: *common
  dbs:
    db2: xyz
`
type Config struct {
    DBs map[string]string `yaml:"dbs"`
}
func main() {
    // Config:  map[common:{DBs:map[db1:abc]} foo:{DBs:map[db1:abc db2:xyz]}]
    var c1 map[string]Config
    err := yaml.Unmarshal([]byte(x), &c1)
    if err != nil {
        fmt.Printf("yaml.Unmarshal[Config]: %#v\n", c1)
    }
    fmt.Printf("Config:\t%+v\n", c1)
    // map: map[common:map[dbs:map[db1:abc]] foo:map[dbs:map[db2:xyz]]]
    var c2 map[string]map[string]map[string]string
    err = yaml.Unmarshal([]byte(x), &c2)
    if err != nil {
        fmt.Printf("yaml.Unmarshal[map]: %#v\n", c2)
    }
    fmt.Printf("map:\t%+v\n", c2)
}
```

This prints:
```
Config:	map[common:{DBs:map[db1:abc]} foo:{DBs:map[db1:abc db2:xyz]}]
map:	map[foo:map[dbs:map[db2:xyz]] common:map[dbs:map[db1:abc]]]
```

I believe the 2nd case is correct; you'd expect the definition of `dbs` on `foo` to override the one pulled in by `<<: *common`.  That matches the output of the ruby YAML parser: `{"common"=>{"dbs"=>{"db1"=>"abc"}}, "foo"=>{"dbs"=>{"db2"=>"xyz"}}}`.

bughou:
yes, i encountered the same problem
```
package main

import (
	"fmt"
	"log"

	yaml "gopkg.in/yaml.v2"
)

func main() {
	content := []byte(`
common: &common
  key:
    a:
      k1: a1
      k2: a2
    b:
      k1: b1
      k2: b2
foo:
  <<: *common
  key:
    a:
      k1: A1
`)
	unmarshalAsMap(content)
	unmarshalAsStruct(content)
}

type m map[string]map[string]string

func unmarshalAsMap(content []byte) {
	var data map[string]map[string]m
	if err := yaml.Unmarshal(content, &data); err != nil {
		log.Panic(err)
	}
[...]

PR #326, opened 2018-02-03
Fix bug in overriden merge in struct (issue #325).

dtbartle:
https://github.com/go-yaml/yaml/issues/325


bughou:
why this fix is not yet merged, any body knowns ?

hans-stripe:
This bug is still an issue for us and we'd like to see this merged.

@niemeyer: since you've been active in PRs recently, would you mind taking a look at this?

PR #329, opened 2018-02-15
Add method to set preferred line width.

### LINKED ISSUE #166: Marshalling long, single-line strings results in (correct) line breaks
When long, single-line strings (i.e. does not contain "\n") are marshalled, the yaml emitter seems to force line breaks at 80 characters, but does so without prefixing the string with the multi-line notation (e.g. ">" or "|"). This results in unusable yaml output like in the example below.

**Example:**

``` go
data := yaml.MapSlice{{"test", "abdasfadsfdasfadsfasd abdasfadsfdasfadsfasd abdasfadsfdasfadsfasd abdasfadsfdasfadsfasd abdasfadsfdasfadsfasd abdasfadsfdasfadsfasd abdasfadsfdasfadsfasd"}}

res, _ := yaml.Marshal(data)
fmt.Println(string(res))
```

**Output:**

``` yaml
test: abdasfadsfdasfadsfasd abdasfadsfdasfadsfasd abdasfadsfdasfadsfasd abdasfadsfdasfadsfasd
  abdasfadsfdasfadsfasd abdasfadsfdasfadsfasd abdasfadsfdasfadsfasd
```

As a workaround, simply appending "\n" to the end of a string that's known to be lengthy works fine, although not ideal since the yaml output now exceeds the 80 char width. Mutiline ">" prefixed outputs would be preferable.

**Output with trailing \n:**

``` yaml
test: |
  abdasfadsfdasfadsfasd abdasfadsfdasfadsfasd abdasfadsfdasfadsfasd abdasfadsfdasfadsfasd aabdasfadsfdasfadsfasd abdasfadsfdasfadsfasd abdasfadsfdasfadsfasd
```

Appreciate if someone could confirm whether this is an issue or just incorrect usage. If a fix is needed, I'd be happy to work on a PR. Thanks!

alexandrevicenzi:
This should fix #166, let me know if I need to change something.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -363,3 +363,7 @@ func (e *encoder) emitScalar(value, anchor, tag string, style yaml_scalar_style_
 	e.must(yaml_scalar_event_initialize(&e.event, []byte(anchor), []byte(tag), []byte(value), implicit, implicit, style))
 	e.emit()
 }
+
+func (e *encoder) setWidth(width int) {
+	yaml_emitter_set_width(&e.emitter, width)
+}
--- a/yaml.go
+++ b/yaml.go
@@ -220,6 +220,13 @@ func NewEncoder(w io.Writer) *Encoder {
 	}
 }
 
+// SetLineWidth sets the preferred line width.
+// To disable long line breaks set width lower than zero.
+// By default, line width is set to 80.
+func (e *Encoder) SetLineWidth(width int) {
+	e.encoder.setWidth(width)
+}
+
 // Encode writes the YAML encoding of v to the stream.
 // If multiple items are encoded to the stream, the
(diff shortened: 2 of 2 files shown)


lox:
Would it be possible to add a test for this @alexandrevicenzi?

alexandrevicenzi:
I'll take a look.

niemeyer:
v3 is coming and we'll use a different style to provide such options. Please feel free to resubmit once it's out.

sagikazarmark:
Is v3 still coming? This seems to be a minor change for a rather serious issue for quite a lot of people, so it would be nice to have this before v3 if it takes another year to reach stable.

niemeyer:
Yep, it's coming.

Issue #184, opened 2016-07-07
MapSlice does not support anchors/merges


Mange:
Hi! I seem to have found a bug in `MapSlice`:

``` go
import (
    "gopkg.in/yaml.v2"
    "reflect"
    "testing"
)

func TestMapMerging(t *testing.T) {
    data := `---
world: &world
  greeting: Hello
earth:
  << : *world`

    config := map[string]interface{}{}
    err := yaml.Unmarshal([]byte(data), &config)
    if err != nil {
        t.Errorf("Failed to parse\n%v\n\ninto a %T: %v", data, config, err)
    }

    earth := config["earth"]
    expected := map[interface{}]interface{}{
        "greeting": "Hello",
    }

    if !reflect.DeepEqual(earth, expected) {
        t.Errorf("Merging does not work. Expected %v (%T), got %v (%T)", expected, expected, earth, earth)
    }
}

func TestMapSliceMerging(t *testing.T) {
    data := `---
world: &world
  greeting: Hello
earth:
  << : *world`

    config := make(yaml.MapSlice, 0)
    err := yaml.Unmarshal([]byte(data), &config)
    if err != nil {
        t.Errorf("Failed to parse\n%v\n\ninto a %T: %v", data, config, err)
    }

    earth := config[1].Value
    expected := yaml.MapSlice{{Key: "greeting", Value: "Hello"}}

    if !reflect.DeepEqual(earth, expected) {
        t.Errorf("Merging does not work. Expected %v (%T), got %v (%T)", expected, expected, earth, earth)
    }
}
```

Output:

```
--- FAIL: TestMapSliceMerging (0.00s)
    foo_test.go:264: Merging does not work. Expected [{greeting Hello}] (yaml.MapSlice), got [] (yaml.MapSlice)
```

Note how the test with the normal `map[string]interface{}` works properly. I'm still a beginner in Go, so I'm not sure where to start to fix this one. I have a really hard time understanding the decoding code, but a small pointer in the right direction might help me find some way of fixing this.

bradrydzewski:
@niemeyer if we send a patch for this what is the process for getting it merged? Is the v2 branch frozen or are you accepting changes? If not, is there a timeline for v3?

cc @donny-dont

vinzenz:
@bradrydzewski could you please try with #220 
This issue might be related to #91

vinzenz:
@bradrydzewski @Mange Its fixed by #220 I took the liberty to add the above test cases to the tests

[...]

Issue #91, opened 2015-03-24
YAML inheritance broken when using yaml.MapSlice


mattes:
See comments in code below:

``` go
package main

import (
    "fmt"
    "gopkg.in/yaml.v2"
)

var body = `
foo: &foo
  some: "var"

bar:
  <<: *foo
`

type Foobar struct {
    Foo interface{}
    Bar interface{}
}

func main() {
    // works as expected
    f := &Foobar{}
    if err := yaml.Unmarshal([]byte(body), &f); err != nil {
        fmt.Println(err)
    }
    fmt.Printf("%+v\n", f) // &{Foo:map[some:var] Bar:map[some:var]}

    // fails parsing yaml inheritance
    f1 := &yaml.MapSlice{}
    if err := yaml.Unmarshal([]byte(body), &f1); err != nil {
        fmt.Println(err)
    }
    fmt.Printf("%+v\n", f1) // &[{Key:foo Value:[{Key:some Value:var}]} {Key:bar Value:[]}]
    // expected: &[{Key:foo Value:[{Key:some Value:var}]} {Key:bar Value:[{Key:some Value:var}]}]
}
```

amolokoedov:
Not yet fixed. Is there a patch for this?

vinzenz:
@mattes @amolokoedov Would be cool if you could check that your use cases are covered with the PR I have sent

rogpeppe:
@vinzenz I cannot find a description of the `<<` operator in any of the YAML 1.x specifications.

Should this actually be a supported feature?

vinzenz:
@rogpeppe http://yaml.org/type/merge.html

rogpeppe:
@vinzenz How many other "working drafts" are actually part of the YAML spec without being mentioned there? How is one supposed to know what YAML actually is without a full specification?

vinzenz:
@rogpeppe Well I don't know, YAML is a beast ;-) I just added the support for MapSlice - Merge support was there already before

hjdr4:
I need to use both merge and override and keep order.
Using https://github.com/vinzenz/yaml/commit/764aa8c1729663c83a7df36e1a68e81ca32ce8f1 I'm able to use merge with MapSlice but overriding a merged key does not work, keys are just duplicated.

niemeyer:
[...]

PR #331, opened 2018-02-16
Fix and test merges with MapSlice

hjdr4:
This patch adds tests and working code when decoding merges into `MapSlice`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+debug.test
--- a/decode.go
+++ b/decode.go
@@ -647,11 +647,21 @@ func (d *decoder) mappingSlice(n *node, out reflect.Value) (good bool) {
 	mapType := d.mapType
 	d.mapType = outt
 
-	var slice []MapItem
-	var l = len(n.children)
-	for i := 0; i < l; i += 2 {
+	var keys []interface{}
+	kv := make(map[interface{}]interface{})
+
+	for i := 0; i < len(n.children); i += 2 {
 		if isMerge(n.children[i]) {
 			d.merge(n.children[i+1], out)
+			for i := 0; i < out.Len(); i++ {
+				item := out.Index(i).Interface().(MapItem)
+				_, ok := kv[item.Key]
+				kv[item.Key] = item.Value
+				if !ok {
+					keys = append(keys, item.Key)
+				}
+			}
+
 			continue
 		}
 		item := MapItem{}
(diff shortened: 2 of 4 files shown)


lox:
Hi @hjdr4! Are you able to resolve the conflicts?

Is this a fix for https://github.com/go-yaml/yaml/issues/91?

hjdr4:
Hello,

I have resolved the conflicts.
I have tested the snippet in https://github.com/go-yaml/yaml/issues/91 and it outputs what is expected.
It also fixes https://github.com/go-yaml/yaml/issues/184.

lox:
Thanks! Could you explain what the issue was? Might this change effect anything else?

hjdr4:
`MapSlice` allows to keep ordering when decoding YAML and this is sometimes required.
The problem is already described in the referenced issues : it does not work correctly at the moment.

So first I wrote a test that shows it's broken (using already defined test data) and code that does what is expected.

I didn't touch the others tests and they are all passing so I guess this won't break anything.

lox:
@rogpeppe any thoughts on this one?

niemeyer:
@lox Any thoughts from you on it? :)

lox:
I'm still getting my head around the impact of the addition of the recursive merge, but it seems to be a clean patch with tests that addresses these specific issues.

lox:
I'm just in the process of testing it.

lox:
👍 from me.

niemeyer:
There are assumptions about typing in the general logic of merge which may not be fulfilled by the structure, in the sense that slices won't necessarily have that type simply because their kind is a slice.

All, I appreciate your help with this, but I'd like to put a tombstone on this issue. This is adding logic to general code paths in v2 which may break down even when _not_ using MapSlice, and I don't want to look back into that. MapSlice will die in v3, and be replaced by a much more flexible intermediate representation. I want to focus on that instead, and anyone that really wants such MapSlices and merges, which is not quite common, can move forward and migrate into it.

Again, thanks for your time here.

lox:
No problems @niemeyer, looking forward to v3!

hjdr4:
Those assumptions are already done https://github.com/go-yaml/yaml/blob/v2/decode.go#L585 and https://github.com/go-yaml/yaml/blob/v2/decode.go#L653.

If you want a stronger test, it's just a matter of changing the merge() test to 
```
out.Kind() == reflect.Slice && out.Type().Elem() == mapItemType 
```

niemeyer:
@hjdr4 The logic in `merge()` itself is general, and called from multiple places. Nothing in there checks the type of the slice.

To be very clear, though, I'm not blaming you and this is a normal issue that we could easily sort out in reviews. But given yaml.v3 is close by and all of that will be dead, I'd indeed prefer to just not touch the logic that is there for several years and instead support any use cases on v3.

Issue #191, opened 2016-08-18
Parsing custom types is not implemented?


ghost:
Suppose I wanted to parse something like this:

``` go
func TestSimpleCustomType(t *testing.T) {
    payload := `

---
myObject:  !myClass { name: Joe, age: 15 }
`
    result := map[interface{}]interface{}{}
    err := yaml.Unmarshal([]byte(payload), &result)
    if err != nil {
        t.Fatalf("Parsing failed: %+v", err)
    }
    fmt.Printf("Parsing succeeded: %+v\n", result)
}
```

(example from Wikipedia)
I'd expect this to tell me that it doesn't know how to unmarshall  `myClass`, instead it produces:

```
Parsing succeeded: map[myObject:map[name:Joe age:15]]
```

By looking into the code of `decode.go` I don't see any attempt to implement this feature. Is this planned?  Is the lack of this ability documented anywhere?

rogpeppe:
With the new Decoder type, we should be able to add support for custom tag decoding...

wryun:
Doesn't this PR implement it?

https://github.com/go-yaml/yaml/pull/204

(another please merge ;) )

niemeyer:
This will be supported in v3 using a custom unmarshaler.

wryun:
It would be nice to have an option which passes through tags unmodified (i.e. can unmarshal/marshal and preserve tags), at least without needing to define custom unmarshallers for every tag you expect to encounter.

niemeyer:
Yes, that will be supported via an intermediate format in v3.

varac:
Ping, any chance to get this fixed ?

PR #332, opened 2018-02-19
Handle custom tags

wryun:
See also https://github.com/go-yaml/yaml/issues/191 and https://github.com/go-yaml/yaml/pull/204

This handles the basic case of 'I want tag information', but doesn't allow one to register custom unmarshalers (yet) outside of that enabled by UnmarshalYAMLWithTag (see addition to yaml.go). However, it seems like this will cover a bunch of use cases, and means that data isn't lost on marshaling/unmarshaling custom tags.

I've also snuck in a backwards incompatible change: validating the existing yaml types.

I'm interested in feedback on the approach. A bunch of tests need to be added, and will be if this is likely to be merged.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -196,6 +196,7 @@ func (p *parser) scalar() *node {
 
 func (p *parser) sequence() *node {
 	n := p.node(sequenceNode)
+	n.tag = string(p.event.tag)
 	p.anchor(n, p.event.anchor)
 	p.expect(yaml_SEQUENCE_START_EVENT)
 	for p.peek() != yaml_SEQUENCE_END_EVENT {
@@ -207,6 +208,7 @@ func (p *parser) sequence() *node {
 
 func (p *parser) mapping() *node {
 	n := p.node(mappingNode)
+	n.tag = string(p.event.tag)
 	p.anchor(n, p.event.anchor)
 	p.expect(yaml_MAPPING_START_EVENT)
 	for p.peek() != yaml_MAPPING_END_EVENT {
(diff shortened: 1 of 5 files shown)


niemeyer:
This will be supported in v3 by unmarshaling into the intermediate representation and then decoding it as desired. We can also consider whether to actually pass the intermediate representation directly into the unmarshal function. Will make porting a bit harder, but everything else more convenient.

wryun:
Do you have an ETA for v3?

wryun:
(or a branch which I can use? There doesn't appear to be a v3 in the repo)

niemeyer:
@wryun You can expect it pretty soon. I'm working on a different project which depends on those new features, which is good both because I'm having proper experience to develop the features, and because it will be coming out soon.

wryun:
@niemeyer any news on this? If not, given that this is backwards compatible with v2, would you reconsider merging it?

varac:
Yes please, I'm waiting for this fix as well...

Snaipe:
Any update on this? I'd rather have that for v2 rather than waiting forever for v3.

niemeyer:
Both cooking changes into v2 or doing a v3 depends on my time, and v3 is mostly there. Moving back to changing v2 now won't make it any faster.

I'm sorry for the delay.

Snaipe:
Hey, no worries -- I ended up [rebasing this PR](https://github.com/Snaipe/go-yaml) onto a forked v2 for the moment. Looking forward to v3! :)

PR #334, opened 2018-02-23
Add note in readme

hazim1093:
As I'm new to go, took me a long time to figure out this issue so thought mentioning here wouldn't be any harm :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,3 +133,6 @@ b:
   - 4
 ```
 
+#### NOTE
+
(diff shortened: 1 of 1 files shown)


niemeyer:
Sorry, but I'd prefer to not have this note on every single Go package. I realize it's a common question, for someone that never used Go, but there will be many other gotchas in those cases and we don't want to turn such third-party packages into the place for those notes.

PR #342, opened 2018-03-12
Convert gocheck to vanilla go library testing

lox:
My first step in exploring this library was to run `go test -v .` on it, which at present doesn't run any of the tests. I assume there is a way to run them with gocheck, but it wasn't immediately obvious from looking at https://labix.org/gocheck or any of the related docs.

Given this library uses very little of the gocheck infrastructure, I thought I'd see what it looked like converted to using the standard `testing` pkg. The upside of this is that it's much easier for new contributors to get involved as it's closer to the "go way".

Given you are the author of gocheck @niemeyer, I imagine you might have different philosophies, by all means feel free to close this if it's either to invasive or not in the direction you are planning!


rogpeppe:
Thanks for this, but I'm not sure this is worth it - it makes the tests longer and less descriptive on failure.
Running `go test -v` does actually run all the tests - it just doesn't print each test as it runs, because all are run within the single function Test. If you run `go test -check.v`, you will see all the tests being run.

lox:
Sure thing! Ended up mostly being an exercise to learn my way around.

PR #353, opened 2018-03-28
Use underlying float precision when formatting floats

### LINKED ISSUE #352: float32 values like 0.99 are formatted incorrectly
While updating to the latest version, I noticed that float32 values like 0.01 and 0.99 are now formatted incorrectly.

E..g, test failure output for 0.99
```
-       weight: 0.99
+       weight: 0.9900000095367432
```

I think this is related to https://github.com/go-yaml/yaml/pull/314

I'm working on a fix.

prashantv:
Currently, all float values are formatted using 64-bit, but that
incorrectly formats float32 values like 0.01 and 0.99.

See https://play.golang.org/p/jbseI1ivyMW for more context.

Fixes #352.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -333,7 +333,13 @@ func (e *encoder) timev(tag string, in reflect.Value) {
 }
 
 func (e *encoder) floatv(tag string, in reflect.Value) {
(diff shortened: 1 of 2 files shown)


niemeyer:
Thanks, LGTM! Let's just wait for tests.

PR #370, opened 2018-06-03
Fix implicit tag resolution

niki5712:
1) According to [spec](http://yaml.org/type/int.html), a value with leading zero is valid only if it's base 8 otherwise it's not an integer, this is confirmed by Regexp:
 ```
 [-+]?0b[0-1_]+ # (base 2)
|[-+]?0[0-7_]+ # (base 8)
|[-+]?(0|[1-9][0-9_]*) # (base 10)
|[-+]?0x[0-9a-fA-F_]+ # (base 16)
|[-+]?[1-9][0-9_]*(:[0-5]?[0-9])+ # (base 60)
```
I suppose we should leave such values as strings and thus we shouldn't quote them in an encoder.

2) I added support binary with '+' sign.

3) I made yamlStyleFloat more strict, now it validates input before parsing it but I still treat very big int as float.

4) Fixed always true condition: `true || intv == int64(int(intv))`.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -25,6 +25,12 @@ var unmarshalTests = []struct {
 	{
 		"{}", &struct{}{},
 	}, {
+		"leading zero: 08765",
+		map[string]string{"leading zero": "08765"},
+	}, {
+		"leading zero: 08765",
+		map[string]interface{}{"leading zero": "08765"},
+	}, {
 		"v: hi",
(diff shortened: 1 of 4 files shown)


niemeyer:
Please have a look at _10.3.2. Tag Resolution_ under the [1.2 specification](http://yaml.org/spec/1.2/spec.html).

The regular expression is simply _[-+]? [0-9]+_.

PR #373, opened 2018-06-13
Allow the DefaultMapType to be overridden

mikefarah:
I'm using this library to decode arbitrary Yaml, perform a transform then output the results (see http://github.com/mikefarah/yq).

The issue I have is that I don't know the root type of the yaml before hand, but I would like to preserve the order of the Map keys if it is a Map (using MapSlice).

This change allows me to override the defaultMapType to MapSlice to that effect.
```
yaml.DefaultMapType = reflect.TypeOf(yaml.MapSlice{})
```

Thoughts?

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -232,16 +232,17 @@ type decoder struct {
 }
 
 var (
-	mapItemType    = reflect.TypeOf(MapItem{})
-	durationType   = reflect.TypeOf(time.Duration(0))
-	defaultMapType = reflect.TypeOf(map[interface{}]interface{}{})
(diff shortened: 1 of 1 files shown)


mikefarah:
:tumble_weed:

virtuald:
I prefer #385 instead of this PR, allowing a [global change to solve a local problem](https://blogs.msdn.microsoft.com/oldnewthing/20081211-00/?p=19873) seems like a bad idea.

mikefarah:
Yep I like that better too, this is quick dirty hack :)

PR #380, opened 2018-07-05
Add Buffered() to Decoder to expose the internal buffer contents

virtuald:
This is useful for tasks such as parsing Jekyll's front matter documents, which can be accomplished now using the following snippet:

```
dec := yaml.NewDecoder(reader)
err = dec.Decode(&value)
... 
remainder, err := ioutil.ReadAll(io.MultiReader(dec.Buffered(), reader))
```

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -3,6 +3,7 @@ package yaml_test
 import (
 	"errors"
 	"io"
+	"io/ioutil"
 	"math"
 	"reflect"
 	"strings"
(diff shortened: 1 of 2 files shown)


virtuald:
An alternative would be implementing a Buffered() method similar to what `encoding/json.Decoder` does, but that seems slightly more annoying to use from a user perspective.

virtuald:
Actually, I found io.MultiReader, which is pretty convenient for this sort of thing. I changed it to `Buffered` instead of `Offset`, to match the json API.

virtuald:
Because this thing uses it's own weird buffering and `yaml_parser_update_buffer` adds extra zeroes to the buffer, extra zeroes end up in the output. I've added a hack to remove them, but it's not clear to me whether this is truly the right solution.

Additionally, because there are two layers of buffering (raw buffer and the parsed buffer), it's possible that if the content after the yaml section has a different encoding than the first section (not sure why you would do that...), then the buffer may yield weird results.

niemeyer:
Thanks for the PR, and sorry for taking a while. It seems reasonable, and I'll get back to it once [v3](https://twitter.com/gniemeyer/status/1073638853225996288) is out.

PR #381, opened 2018-07-05
Allow specifying flow or block style at encode time

benesch:
It is often useful to be able to encode a given value in both block and
flow style. The current struct tag-based API requires committing to
either block or flow style for a given field. It also provides no means
of requesting that the top-level struct use flow style.

Introduce a new API, Encoder.UseStyle, which requests that all encoded
values use a particular encoding style.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -17,6 +17,7 @@ type encoder struct {
 	emitter yaml_emitter_t
 	event   yaml_event_t
 	out     []byte
+	style   EncodingStyle
 	flow    bool
 	// doneInit holds whether the initial stream_start_event has been
 	// emitted.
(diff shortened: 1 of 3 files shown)


knz:
@niemeyer do you reckon this could make it to a go-yaml release before october?

niemeyer:
Not as it is.. but I do hope that before October we'll have a release of go-yaml v3, that will allow you to implement those changes on top of the stock API easily, by having access to a lower level representation. I'm not that far, and am hoping to have some more time to finish it soon.

knz:
thank you

niemeyer:
Taking a little longer than expected, but v3 is coming. Here is a sneak peek:

https://twitter.com/gniemeyer/status/1073638853225996288

Style may be arbitrarily tuned for any of the nodes.

Closing this for now.

PR #382, opened 2018-07-06
Allow Marshal / UnMarshal of reflect.Value objects

AnthonyPoschen:
this change solves the problem of having a reflect.Value that you wish to send to unmarshal/marshal and get meaningful / expected behavior for advanced usage cases.

it makes it super clean and avoids needing to send an empty interface and parsing the resulting map[interface{}]interface{} as it directly sets against the reflect.value passed in.

i am happy to explain my use case in detail for wanting this change if desired.

### DIFF
--- a/yaml.go
+++ b/yaml.go
@@ -142,6 +142,9 @@ func unmarshal(in []byte, out interface{}, strict bool) (err error) {
 	node := p.parse()
 	if node != nil {
 		v := reflect.ValueOf(out)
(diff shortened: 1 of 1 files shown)


niemeyer:
With a reflect.Value at hand, it should be possible to just say value.Interface() to get something that may be handled into encoding or decoding.

If that doesn't work for your use case, I'd certainly be curious to hear what it is.

Tentatively closing for now.

Issue #384, opened 2018-07-09
Better support for maps with string keys


algobardo:
The following code put into `out` a `map[interface{}]interface{}` so preventing any `json.Marshal` of the resulting object. Shouldn't the resulting type be map[string]interface{}?

```
func Test__yaml_unmarshal_handle_empty_maps_with_right_type(t *testing.T) {
	var out interface{}
	err := yaml.Unmarshal([]byte("{}"), &out)
	require.NoError(t, err)
	require.Equal(t, reflect.TypeOf(out).String(), "map[string]interface{}")
}
```

virtuald:
It seems that from a backwards compatibility stance, changing the default map type probably isn't an option for 2.x. However, #385 is a great solution for that.

niemeyer:
In v3, decoding a map into an interface{} when it has _only_ string keys will decode as map[string]interface{}, and not map[interface{}]interface{}. If there are non-string keys, the latter will be used to ensure the values can roundtrip.

niemeyer:
I'm hoping this will properly address the mentioned cases without breaking usability for general yaml maps. I will close this issue once v3 is generally available.

nikolaas:
Hello. Unfortunately, your solution won't work with the case when we have no chance to change structure of a YAML file. For example, we need to deserialize an OpenAPI v2/v3 YAML file as `map[string]interface{}` but it contains HTTP status codes as keys:
```yaml
openapi: "3.0.0"
paths:
  /persons:
    operationId: Returns a person's list
    responses:
      200:
        description: A person's list
      500:
        description: An internal server error
```

What options do we have to resolve the problem?

Thanks.

PR #385, opened 2018-07-09
Added ability to customize default map type

algobardo:
Would allow me to cope with #384

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -240,8 +240,11 @@ var (
 	ptrTimeType    = reflect.TypeOf(&time.Time{})
 )
 
-func newDecoder(strict bool) *decoder {
-	d := &decoder{mapType: defaultMapType, strict: strict}
+func newDecoder(strict bool, mapType reflect.Type) *decoder {
+	if mapType == nil {
+		mapType = defaultMapType
+	}
+	d := &decoder{mapType: mapType, strict: strict}
 	d.aliases = make(map[*node]bool)
 	return d
 }
--- a/decode_test.go
+++ b/decode_test.go
@@ -1296,6 +1296,7 @@ func (s *S) TestFuzzCrashers(c *C) {
 	}
 }
 
+
 //var data []byte
 //func init() {
 //	var err error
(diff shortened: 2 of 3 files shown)


steve-gray:
+1 on this. The inability to do this breaks a *lot* of code that works on maps, and requires a lot of fairly evil hacking to work around (recursive replacement of maps)

mikefarah:
Nice - do you mind giving a little snippet of how this change could be used to change the default map type?

virtuald:
Something like

```
decoder := NewDecoder()
decoder.SetDefaultMapType(reflect.TypeOf(map[string]interface{}{}))
err := decoder.Unmarshal(whatever, thing)
```

dlahoza:
+1 from me

niemeyer:
I've been working on v3, and will consider the underlying problem here before releasing to make sure it is addressed. Without a new version where we can break compatibility in edge cases, I think we may be able to address this issue without an explicit function call like this, by recording the generic map the decoder last went through. I've done the same in a different marshaling package and it worked well. I'll update this issue once v3 is ready for testing.

niemeyer:
Okay, in v3 decoding a map into an interface{} when it has _only_ string keys will decode as map[string]interface{}, and not map[interface{}]interface{}. If there are non-string keys, the latter will be used to ensure the values can roundtrip.

I'm hoping this will properly address the mentioned cases without breaking usability for general yaml maps. I will close this pull request now, and will close the related issue once v3 is generally available.

ktogo:
@niemeyer In the meantime I would still love this solution to be merged, so that we can move forward for now.

Also even on v3 we’ll be able to lock the type into `map[string]interface{}` where we don’t expect other type as map keys with this default type hinting solution.
(In that case developers need to implement the type casting/validating code at everywhere they need string type keys without it)

marwatk:
@niemeyer Would it be possible to reexamine this MR? It's pretty minor and it's been a year since the closure and v2 is still the default branch. A lot of projects are using a fork to accomplish it, but we'd rather not have to.

PR #389, opened 2018-07-24
do not decode {} as nil MapSlice

jdhenke:


jdhenke:
Before making the fix to `decode.go`, this test fails by unmarshaling `{}` as `nil`.

```
$ go test

----------------------------------------------------------------------
FAIL: decode_test.go:1299: S.TestNilMapSlice

decode_test.go:1320:
    c.Errorf("Unmarshaling %s as a MapSlice incorrectly yielded a %s MapSlice", tc.yaml, result)
... Error: Unmarshaling {} as a MapSlice incorrectly yielded a nil MapSlice

OOPS: 31 passed, 1 FAILED
--- FAIL: Test (0.02s)
FAIL
exit status 1
FAIL	gopkg.in/yaml.v2	0.177s
```

Happy to fixup style and fix as needed, just let me know. Cheers.

jdhenke:
Besides a specific use case where this behavior caused problems, it seems odd that unmarshaling into a regular `map` would have a different `nil`-behavior than a `MapSlice`, which it currently does. 

```
$ cat main.go
```

```go
package main

import (
	"fmt"

	yaml "gopkg.in/yaml.v2"
)

func main() {

	b := []byte("{}")
	var m map[string]string
	if err := yaml.Unmarshal(b, &m); err != nil {
		panic(err)
	}
	fmt.Printf("Map: %#v Nil: %v\n", m, m == nil)
	var ms yaml.MapSlice
	if err := yaml.Unmarshal(b, &ms); err != nil {
		panic(err)
	}
	fmt.Printf("MapSlice: %#v Nil: %v\n", ms, ms == nil)
}
```
```
$ go run main.go
Map: map[string]string{} Nil: false
MapSlice: yaml.MapSlice(nil) Nil: true
```

niemeyer:
The most common form of an empty slice is nil, and all common operations, including mutation, work fine on it. So it's not clear what is the actual issue being fixed here.

Either way, MapSlice is going away in v3, and is being replaced by a much richer intermediate state, so it seems better not to touch this now.

Thanks for offering the fix, though. It might easily be an actual bug.

PR #399, opened 2018-08-30
Update tested golang versions

meatballhat:
using (recently added) version alias support and `master` instead of legacy `tip` alias.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,14 @@
 language: go
 
 go:
-    - 1.4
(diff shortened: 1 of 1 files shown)


thetillhoff:
The versions were already added in [https://github.com/go-yaml/yaml/commit/1ed59511881fdb008c1e618e9f219ce0704e658e](https://github.com/go-yaml/yaml/commit/1ed59511881fdb008c1e618e9f219ce0704e658e)

PR #413, opened 2018-11-05
Able to skip exported fields

norshtein:
This PR makes `goyaml` able to skip exported fields in struct. To skip marshaling and unmarshaling exported fields, just add a `skip` tag in struct tags.


niemeyer:
The common convention for that is "-", also supported by the yaml package.

PR #414, opened 2018-11-05
Add support for json.Number.

echlebek:
This commit adds support for encoding json.Number as a number
instead of as a string, which is the underlying type of
json.Number.

Care has been taken to not introduce a dependency on the
encoding/json package, by using an interface that specifies the
methods of json.Number instead of the datatype itself. This also
means that other packages that use a similar datatype will be
supported. (Like json-iterator)

This is useful for tools that wish to convert JSON data into YAML
data by deserializing JSON into a map[string]interface{}, and use
the json.Encoder's UseNumber() method, or structs that use the
json.Number datatype and also wish to support yaml.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -714,6 +714,14 @@ var unmarshalTests = []struct {
 		"---\nhello\n...\n}not yaml",
 		"hello",
 	},
+	{
+		"a: 5\n",
+		&struct{ A jsonNumberT }{"5"},
+	},
+	{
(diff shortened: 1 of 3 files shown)


echlebek:
I based this work on v2, since it's the version I'm using, but I'm happy to base it on the devel branch if that would be more appropriate.

niemeyer:
Thanks!

echlebek:
Thank you!

PR #419, opened 2018-12-10
Preserve comments

### LINKED ISSUE #132: Support (un)marshaling of comments

genevieveluyt:
Fixes #132 

This PR adds support for parsing and emitting yaml with comments. This change is intended to be completely backward compatible.

Comments are only supported when using a MapSlice as the output object because a MapSlice preserves the order of map items. This means that output options which can normally be specified in a struct are not supported. The output will use defaults for everything (eg. flow maps will be converted to a list of dashed items instead of an array format)

ie. this is not supported
```golang
type T struct {
	B struct {
		RenamedC int   `yaml:"c"`
		D        []int `yaml:",flow"`
	}
	A string
}
t := T{}
```

instead you would use simply
```golang
m := yaml.MapSlice{}
```

## Example
The input
```
# pre doc comment 1
# pre doc comment 2

%YAML   1.1

%TAG    !   !foo
%TAG    !yaml!  tag:yaml.org,2002:

---

# pre doc comment 3
b: # map key comment
  # first line comment
  c: 2 # end of line comment
  # flow leading comment
  d: [3, 4] # flow eol comment
  # comment 1
  e: # sequence key comment
    # comment 2
    - 3 # seq eol comment 1
    - 4 # seq eol comment 2
  # last line comment
# comment 3
a: Easy!
# comment 4
? e # key eol comment
: 8 # value eol comment
# post doc comment

# post doc comment 2
```

results in the output
```
# pre doc comment 1
# pre doc comment 2

%YAML   1.1

%TAG    !   !foo
%TAG    !yaml!  tag:yaml.org,2002:

---

# pre doc comment 3
b: # map key comment
  # first line comment
  c: 2 # end of line comment
  # flow leading comment
  # flow eol comment
  d:
  - 3
  - 4
  # comment 1
  e: # sequence key comment
  # comment 2
  - 3 # seq eol comment 1
  - 4 # seq eol comment 2
# last line comment
# comment 3
a: Easy!
# comment 4
e: 8 # key eol comment; value eol comment
# post doc comment
# post doc comment 2
```

### Differences:

image

### Leading Comments
Comments and directives at the top of the yaml document (anything before data starts) are copied straight to the output. That means they preserve all formatting, including newlines. For the previous example, the leading comment is
```
# pre doc comment 1
# pre doc comment 2

%YAML   1.1

%TAG    !   !foo
%TAG    !yaml!  tag:yaml.org,2002:

---

# pre doc comment 3
```

### End of Line Comments (EOL Comments)
When using the comment marshaler, all map and sequence items are structs with a `Comment` field. For reverse compatibility, this is *only* the case when using the comment unmarshaler.

Eg.
```
a: # comment 1
  b: 3 # comment 2
  c:
    - 2 # comment 3
```
results in an identical output.

End of line comments for a flow sequence are put before that sequence.
```
a:
  b: [1, 2] # comment
```
becomes
```
a:
  b:
    # comment
    - 1
    - 2
```

When both an explicit key and its value have an end of line comment, they are combined
```
? a # comment 1
: 6 # comment 2
```
becomes
```
a: 6 # comment 1; comment 2
```

### Single Line Comments
Comments that are on their own line are assumed to describe the following line. When the input yaml is converted into a tree, these comments have the same parent as whatever data line follows them, unless they are at the end of the document, in which case their parent is the root. This means they have the same indentation in the output as the following line.

Eg.
```
# a for apple
a:
  # b for banana
  b:
    # c for cat
    - c
```
results in an identical output.

Using the indentation for the following line does not always work, for example
```
a:
  b:
    - 2
    #- 3
c: 5
```
becomes
```
a:
  b:
    - 2
#- 3
c: 5
```

However copying the comment including its leading space (which is what https://pypi.org/project/ruamel.yaml/ does) could result in too much indentation because indentation is automatically determined by the library
```
a:
       # comment
       - 2
```
becomes
```
a:
       # comment
- 2
```

### Post Doc Comments
These are comments which take place after the yaml data. They are treated the same as map item comments on the root map node (all formatting is discarded) because determining whether data continues after the comments or not would require a look-ahead.

### Literal Style
Literal style and folded strings are not yet supported. Right now, the library simplifies literal style based on the chomping indicator, but this does not make sense when comments are preserved.

For example
```
a: |-
        hello

   # comment
```
cannot be converted to use just a `|` instead of a `|-` while still preserving comments and representing the data accurately. I believe the ideal output here is to retain the `|-` but I would like feedback on this approach and its possible implementation strategy.

image


genevieveluyt:
@niemeyer Could you please explain the CI setup? It looks like the tests are being applied to `gopkg.in/yaml.v2` but since that does not include my changes, the tests are failing. I am not sure how to make the build pass?

niemeyer:
We use the [go_import_path](https://docs.travis-ci.com/user/languages/go/#go-import-path) setting for Travis, which means the code tested should indeed be the one in the repository, despite the vanity import path.

Thanks for your contributions, but please note that I've been working on a new v3 version which I hope to release this year still, which will render these changes incompatible. There I've also been working on initial support for comments, but given what I see in your comments above it will likely not be as complete as your PR seems to be. On the bright side, the main change in v3 is the exposure of an intermediate state, which is much more appropriate for preserving comments (and style!) than what we currently have in v2. I'm currently in the final stages of finishing that (testing and last fixes mainly) before exposing an initial version for everyone to evaluate.

Might be best to pause a bit this work, and re-engage once v3 is ready for testing in a few days. May I have your email, in case we need to communicate elsewhere?

genevieveluyt:
@niemeyer thanks for getting back to me. Glad to hear v3 is near! I will send you my email.

PR #426, opened 2019-01-07
Accept !!env tag in yaml file

mcorbin:
This commit adds support for a `!!env` tag, which can be used to load
a variable from an environment variable.

For example:

```
host: !!env HOSTNAME
```

will load the host value from the HOSTNAME environment variable.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"os"
 	"reflect"
 	"strconv"
 	"time"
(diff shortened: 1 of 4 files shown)


niemeyer:
We can't do that, as it means people can ship configuration files that will blindly load data from the system's machine, which can be exploited in several ways. It's also an abuse of the "tag:yaml.org,2002" namespace which the "!!" shortcut refers to, but that's minor in comparison and might be sorted by using a different namespace.

The good news is that upcoming v3 will allow you to have a look at such tags and decode them yourself if you really want to. 

Thanks for collaborating and submitting your work, though.

mcorbin:
Thanks you for your quick answer, I will wait for the v3 ;)

Issue #410, opened 2018-11-01
Merge with overrides doesn't work in strict mode


juzna:
When you merge dicts, it's allowed to override keys (http://yaml.org/type/merge.html).

Example 1:
```yaml
foo: &foo
  some: foo

bar:
  <<: *foo
  some: bar # override for `some`, should be allowed
```

Example 2:
```yaml
foo: &foo
  some: foo

bar: &bar
  some: bar

foobar:
  <<: [*foo, *bar]  # override for `some`, should be allowed
```

However, when using the strict mode (`unmarshalStrict`), you get the following error:
`key "some" already set in map`

juzna:
Note to self: this was introduced in 5edc3ded41385ca1b9a80339d2a070e4d0a17cb6 which started enforcing key uniqueness in strict mode. I think that code needs to be improved to correctly support overrides when merging.

andreynering:
I also faced this issue on https://github.com/go-task/task/issues/112 and had to give up on using the strict mode, for now, because of it.

niemeyer:
In v3 we will enforce unique keys at all times, and this problem is also fixed. Will close once v3 is widely available.

6543:
duplicate of #624 ?

blade2005:
Any updates on this issue @andreynering 

v3 is out (v3.30.1) and I'm having this issue.

PR #431, opened 2019-01-20
Fix merge with overrides in strict mode 

b1zzu:
This pull request will fix this issue: #410

Values from a mapping node are put temporary reflect map (and also check for duplicated keys) and finally the temporary reflect map is merged (overriding existing keys) inside the out reflect map (which already contain keys and values from the merged mapping nodes)


niemeyer:
This is fixed in v3 in a slightly different way, so that we don't set the map twice and also so we get proper debugging information with references to both definitions. It also drops the mention of types in the error message, which is something awkward to surface to users.

PR #442, opened 2019-03-11
Fix -0b on 32-bit systems

### LINKED ISSUE #441: -0b broken on 32-bit systems
https://github.com/go-yaml/yaml/commit/7b8fd2d seems broken on 32-bit systems, in two ways.

Firstly the decode test case:
```
"bin: -0b1000000000000000000000000000000000000000000000000000000000000000",
map[string]interface{}{"bin": -9223372036854775808},
```
won't pass on a system where Int is 32 bits wide because the interface expects an Int, so you get a constant overflow. I fixed that (not realising this edge case was for #87) by changing to:
```
"bin: -0b10000000000000000000000000000000",
map[string]interface{}{"bin": math.MinInt32},
```

However I then ran into a real test failure with this test:
```
                "int64_neg_base2: -0b111111111111111111111111111111111111111111111111111111111111111",
                map[string]int64{"int64_neg_base2": -math.MaxInt64},
```
Which gave me: `["int64_neg_base2"]: 1 != -9223372036854775807`

I think by hardcoding `true ||`, the fix for #87 encodes the assumption that any Int64 value can fit into an Int, which is not the case on a 32-bit system. Deleting the `true ||` fixed this test for me.

ramcq:
7b8fd2d introduced a regression on 32-bit systems where -0b values would be
cast to an Int before being returned as a Int64, meaning that on 32-bit
systems, any negative value smaller than MinInt32 would get wrapped around and
returned incorrectly, causing the int64_neg_base2 decode_test to fail.

The test case also causes the bin test to fail on 32-bit systems because it
presumes the 64-bit edge case value can fit within an Int, which is not the
case on a 32-bit system.

Move the newly-introduced test case to the Int64-specific test cases, and
remove the undesirable coersion of Int64 values to Int when they do not fit.

Fixes #441.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -130,9 +130,6 @@ var unmarshalTests = []struct {
 		"bin: -0b101010",
 		map[string]interface{}{"bin": -42},
 	}, {
-		"bin: -0b1000000000000000000000000000000000000000000000000000000000000000",
-		map[string]interface{}{"bin": -9223372036854775808},
-	}, {
 		"decimal: +685_230",
 		map[string]int{"decimal": 685230},
 	},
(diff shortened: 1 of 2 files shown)


sarg:
@niemeyer could that be merged? It prevents building i686 version of this package in guix: https://ci.guix.gnu.org/build/426173/details

PR #443, opened 2019-03-22
Fix decoding of quoted scalars "null" and "~"

aleksey-mashanov:
Quoted scalars "null" and "~" should be handled the same way
as any other strings which includes calling UnmarshalYAML method.


niemeyer:
Thanks for the fix. Can we please have a test covering the issue, and would you mind to sign the [contributor agreement](http://www.ubuntu.com/legal/contributors)?

aleksey-mashanov:
Surely, done.

aleksey-mashanov:
@niemeyer, any progress on merging this?
I have signed contributor agreement and wrote tests.

PR #449, opened 2019-04-04
Set correct module path

andrewheberle:
Updates the "go.mod" file so the module path is correct for the v3 branch


niemeyer:
Thanks!

Issue #458, opened 2019-05-06
v3 RFE: export ending line/column in nodes, add StartIndex and EndIndex


ajeddeloh:
I'm thrilled that v3 has the Node type exported, especially with the Line and Column fields. It would be good to expose the Lines/Column of the end of each node as well. I'm going to be using this in a project where having the upper/lower bounds would be useful and allow better error reporting. Additionally, it would be nice to have the index (i.e. absolute index in bytes or characters) for the start and end of nodes as well.

PR #461, opened 2019-05-08
node: add more data about lines, columns, indexes

ajeddeloh:
Add Index, LineEnd, ColumnEnd, and IndexEnd to Node. *End mark the end
of where that node occurs in the source. Index* marks the index into the
input stream.

Fixes https://github.com/go-yaml/yaml/issues/458

Edit: Needs test updates, will push another commit with those


atomatt:
#593 borrows a bit of this (just for scalars, see description for explanation) and adds doc comments and tests.

PR #485, opened 2019-07-15
travis should use the v3 import path not v2

laverya:
otherwise tests of functions that do not exist in v2 fail

EDIT:
Also, without this pr tests are run against the current v3 code, not the code in the pr
This largely invalidates the point of CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,4 +12,4 @@ go:
     - "1.12"
     - tip
 
-go_import_path: gopkg.in/yaml.v2
+go_import_path: gopkg.in/yaml.v3


laverya:
@niemeyer if the project is closed to PRs, just let me know
But leaving every pr here open without response is really depressing, especially things that should be approvable at a glance ☹️

niemeyer:
Every PR is open because you are only looking at open PRs.

I appreciate your help, and your patience.

laverya:
It's been more than three months since your last comment on an open pr / merge of a pr / etc
That's what I was thinking of 😢

Thank you for the work you do on this project!

laverya:
@niemeyer I think I should point out the importance of this PR

Without it, TravisCI uses the submitted code as `gopkg.in/yaml.v2`, which means that unit tests will fetch the v3 package from github to run against. This means you can't use the unit tests as evidence for a PR not breaking things...

(and for this reason I've merged this branch into most of my PRs - they all have tests that rely on updated behavior, and so would fail otherwise)

PR #488, opened 2019-07-19
chore (v3): properly formatted go.mod and added go.sum

laverya:
this was done with 'go mod tidy'

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,3 @@
-module "gopkg.in/yaml.v3"
+module gopkg.in/yaml.v3
 
-require (
-	"gopkg.in/check.v1" v0.0.0-20161208181325-20d25e280405
-)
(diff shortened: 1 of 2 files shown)


janisz:
What's the status of this? Without `go.sum` @dependabot is not able to verify dependencies

msugakov:
This and https://github.com/go-yaml/yaml/pull/539 seem to be the same change. Without either of these merged, github's @dependabot is not able to verify dependencies of our code that depends on go-yaml.
Having `go.sum` is simply beneficial for reproducible builds and package maintenance tools.
It seems that dependencies of go-yaml have not changed until today so it should be very simple to merge this PR.
Is it possible to merge and release it any time soon @niemeyer?
Thanks!

SVilgelm:
@niemeyer Could you please look at this? Please update and merge, we really need the `go.sum` file to properly maintain the dependencies

dolmen (approved):

PR #490, opened 2019-07-22
Bugfix (v3): Marshal yaml 1.1 bool strings as explicit strings

### LINKED ISSUE #489: Yaml 1.1 bool string values should be quoted
When encoding a string like `yes` or `Y` or `On`, we should include quotes to preserve compatibility with the [yaml 1.1 spec](https://yaml.org/type/bool.html). This can lead to bugs where the generated yaml is incompatible with tools like kustomize, as discovered in this ship issue https://github.com/replicatedhq/ship/issues/1035.

laverya:
String values matching a boolean literal from the [yaml 1.1 spec](https://yaml.org/type/bool.html) should be rendered with quotes to ensure backwards compatibility. This fixes #489.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -627,12 +627,9 @@ func (d *decoder) scalar(n *Node, out reflect.Value) bool {
 		case string:
 			// This offers some compatibility with the 1.1 spec (https://yaml.org/type/bool.html).
 			// It only works if explicitly attempting to unmarshal into a typed bool value.
-			switch resolved {
-			case "y", "Y", "yes", "Yes", "YES", "on", "On", "ON":
-				out.SetBool(true)
-				return true
-			case "n", "N", "no", "No", "NO", "off", "Off", "OFF":
-				out.SetBool(false)
+			val, found := boolMap[resolved]
+			if found {
+				out.SetBool(val)
 				return true
 			}
 		}
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,4 +12,4 @@ go:
     - "1.12"
     - tip
 
-go_import_path: gopkg.in/yaml.v2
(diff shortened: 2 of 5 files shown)


laverya:
Well that's interesting
Running this locally worked for me, but TravisCI disagrees (except with the latest golang?)

I'll have to look into that

laverya:
Seems it was the CI import path
Travis was running the test cases against the current v3 code, not against the code in the PR
Another reason to really require merging #485 soon, I suppose

ppmathis:
@niemeyer Could you please consider merging this PR? I'm currently using go-yaml v3 to unmarshal a YAML file, make some changes to a few structs and marshal it back into YAML. Due to this specific bug, `test: 'yes'` turns into `test: yes`, which therefore changes the type from string to boolean.

laverya:
this also fixes https://github.com/go-yaml/yaml/issues/214, which the issue I opened is a dupe of...

laverya:
Obsoleted by #583 .

decode.go
```diff
@@ -632,12 +632,9 @@ func (d *decoder) scalar(n *Node, out reflect.Value) bool {
 		case string:
 			// This offers some compatibility with the 1.1 spec (https://yaml.org/type/bool.html).
 			// It only works if explicitly attempting to unmarshal into a typed bool value.
-			switch resolved {
-			case "y", "Y", "yes", "Yes", "YES", "on", "On", "ON":
-				out.SetBool(true)
-				return true
-			case "n", "N", "no", "No", "NO", "off", "Off", "OFF":
-				out.SetBool(false)
+			val, found := boolMap[resolved]
```
Ark-kun, line 635: Will this still decode `y` as true?

PR #491, opened 2019-07-31
use double quotes for string containing only equal sign

wgerlach:
triggers yaml_DOUBLE_QUOTED_SCALAR_STYLE if string contains only equal character


wgerlach:
I have not found a human-readable description in the YAML specification when a string needs to be double quoted and when not. For pointers I would be thankful. My pull request handles only the special case `key: =` vs `key: "="`, but I'd be happy to include others cases as well.

PR #507, opened 2019-09-18
FIX: isStringMap should just check the keys of map

PChou:
i'm not sure the fix is on the right way, but it works in my situation.
i think your old code check the key and also the value, which may failed to use `string map` when some value are not `!!str`, even all keys of map are `!!str`.


niemeyer:
Good catch, thanks for the fix.

PChou:
opps! i am trying to fix the test

niemeyer:
Thanks again.

PR #510, opened 2019-09-29
Upgrade to go v1.13

chenrui333:


dolmen:
Why?

chenrui333:
@dolmen, why not?

niemeyer:
We want to make sure it continues to work in prior versions. Adding to the list is fine, though.

laverya:
We should probably also compile with `1.10`-`1.12`, just for completeness.

chenrui333:
That makes sense, I will add more golang versions for testing.

chenrui333:
@niemeyer I think it should be good now!

.travis.yml
```diff
@@ -1,12 +1,26 @@
...
+  - "1.9.x"
+  - "1.10.x"
+  - "1.11.x"
+  - "1.12.x"
+  - "1.13.x"
+  - tip
+
+os:
+  - linux
+  - osx
```
niemeyer: Can we please restrict this PR to this content only. The rest is unrelated to what the PR proposes and I'd rather get the main intention in right now than keep discussing unrelated bits here.

niemeyer (changes requested):

.travis.yml
chenrui333: Sounds good, I will refine the travis build changes.

chenrui333: @niemeyer Updated, now the syntax should be using the latest versions of each golang lines.

laverya: @chenrui333 I think what was meant by that was to not include the go.mod/go.sum changes in this PR

chenrui333: Interesting, so we just need travis build config changes?

laverya: Presumably, since the pr into the v3 branch I made doesn't touch the go.mod/go.sum files, and they're the same between the branches

niemeyer:
I went ahead and did the change simply adding the missing versions on 1ed59511.

chenrui333:
Thanks @niemeyer.

Also, I just create a new PR for cleanup `go.mod` and `go.sum`.

PR #455, opened 2019-04-23
Feature (v3): Add a function to set the desired line length of an Encoder

laverya:
This PR exposes a function `func (e *Encoder) SetLineLength(characters int)` that allows setting the line wrap length of an encoder. The interface was intended to have a very similar style to the SetIndent function `func (e *Encoder) SetIndent(spaces int)`, implemented [here](https://github.com/go-yaml/yaml/commit/2c9fce8bede05098317e715fe41cc2db4847ea0b).

Internally, it uses the existing function `func yaml_emitter_set_width(emitter *yaml_emitter_t, width int)`, which was previously unused.

This relates to several issues (#387, #348, #166) and a previous PR (#329).

PR #524, opened 2019-10-17
add Encoder.SetLineLength API and change imports

jmhodges:
This adds the SetLineLength method to Encoder and changes the imports to the
jmhodges/yaml.v2 config.

The SetLineLength patch is adapated from
https://github.com/go-yaml/yaml/pull/455

The README got an update for folks needing this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,4 @@ go:
     - 1.9
     - tip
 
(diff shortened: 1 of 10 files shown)


jmhodges:
Sorry, wrong repo!

PR #525, opened 2019-10-17
test gopkg.in/yaml.v3, not v2, in limit_test.go

laverya:
otherwise the test is pretty useless
(I noticed because `go mod tidy` was adding v2 as a dependency, which seemed very wrong)


niemeyer:
Thanks!

PR #526, opened 2019-10-17
test go versions up to and including 1.13 on the v3 branch

laverya:
also test on both osx and linux

copies https://github.com/go-yaml/yaml/pull/510, but for the v3 branch


chenrui333:
Nice!

laverya:
The only thing I'm worrying about is the osx builds
Since those seem to have much longer queue times than linux builds do

But I don't know if that's just a "we caught Travis at a bad time" thing

laverya:
It's not looking like a one-time thing
I'm going to drop osx for all but 1.13
best of both worlds that way (validation that it works on osx, without each CI run taking 3+ hours)

niemeyer:
I went ahead and did the change that exclusively adds 1.13 on 196cff6e.

PR #535, opened 2019-11-02
Proposal: Support repeated JSON documents

msteffen:
Hello! I hope this PR-out-of-the-blue is welcome; please let me know if there's more I can do to introduce myself or improve the change.

This makes a modification to the `go-yaml` parser that allows it to parse repeated flow maps similarly to `encoding/json`.

The YAML syntax this change supports is not valid per the YAML 1.2 spec (at least as verified by https://hackage.haskell.org/package/YamlReference), but extends the grammer only slightly beyond the official spec, if I understand the parser correctly, and only to support syntax already parsed elsewhere in the Go ecosystem.

The main advantage of merging this change is that it would allow users of `go-yaml` to completely replace their use of `encoding/json` with `go-yaml` (as Pachyderm aims to do).

Thanks for considering this change, and please let me know if there's any other help or information I can provide!

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -886,6 +886,14 @@ var decoderTests = []struct {
 		"hello",
 		"goodbye",
 	},
+}, {
+	// Parse repeated flow maps as separate documents, for compatibility with Go's
+	// encoding/json
+	"{ \"a\": \"b\" }\n{ \"c\": \"d\" }",
+	[]interface{}{
+		map[string]interface{}{"a": "b"},
+		map[string]interface{}{"c": "d"},
+	},
 }}
 
 func (s *S) TestDecoder(c *C) {
--- a/parserc.go
+++ b/parserc.go
@@ -321,7 +321,7 @@ func yaml_parser_parse_document_start(parser *yaml_parser_t, event *yaml_event_t
 		if token == nil {
 			return false
 		}
-		if token.typ != yaml_DOCUMENT_START_TOKEN {
+		if token.typ != yaml_DOCUMENT_START_TOKEN && token.typ != yaml_FLOW_MAPPING_START_TOKEN {
 			yaml_parser_set_parser_error(parser,
 				"did not find expected <document start>", token.start_mark)
 			return false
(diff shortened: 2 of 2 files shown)


msteffen:
This PR is a bit old at this point, but I just rebased it onto the upstream head. I'm adding @niemeyer as a reviewer in case that was a necessary but missing part of the PR process--apologies for the noise if not!

Let me know if you'd be interested in merging this PR. I'll close it if not

niemeyer:
Hi @msteffen,

That's a very interesting timing for this PR. I was just looking at issue #614 reported recently as well.  Do you want to express an opinion there?

msteffen:
@niemeyer I wrote a response on #614 and then decided to post it here instead, in case it was too off-topic to be useful (thought I could try to investigate that issue if it would be helpful), but my opinion on it is that only one specific class of invalid yaml be parseable by go-yaml, which is repeated JSON documents of the form:
```
{
  "doc": "one"
}
{
  "doc": "two"
}
```
These are parsed by `encoding/json`, and if they were parsed by go-yaml as well, it would be possible to replace `encoding/json` with go-yaml everywhere. As things are, this replacement can't be done without invalidating some inputs.

This raises the question of how the following document (similar to the one in #614) should be parsed:
```
{
  "doc": "one"
}
doc: two
```

My relatively uninformed opinion is that that this document should be invalid, and that the transition function  should look something like
```
"}" -> "{" => end flow document and begin new flow document
"}" -> "^---$" => end flow document and begin new yaml document
"}" -> * => invalid
```
I think this transition function allows repeated documents similar to the one up top while minimally expanding the grammar. It would require the second document above to be rendered using the more intuitive (to me) syntax:
```
{
  "doc": "one"
}
---
doc: two
```

However, I think the change pointed out in #614 out is different; IIUC, go-yaml isn't even trying to parse `this is not yaml` (which isn't yaml) and is just discarding that text. I think that may indeed be a bug, but I think discarding that yaml is compatible with the behavior proposed here.

PR #539, opened 2019-11-05
chore(v2): cleanup go.mod and add go.sum

chenrui333:
relates to #510

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,5 @@
-module "gopkg.in/yaml.v2"
+module gopkg.in/yaml.v2
 
-require (
-	"gopkg.in/check.v1" v0.0.0-20161208181325-20d25e280405
-)
+go 1.13
+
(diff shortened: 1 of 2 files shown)


niemeyer:
I can't see a reason to hardcode the version of go-check. We want whatever is the latest version of v1 in our tests, not that specific digest.

chenrui333:
@niemeyer I think the intention of `go.sum` is good for cross-module reference. The `go.mod` is just a result of `go mod tidy` though.

janisz:
@niemeyer What's the status of this? Without `go.sum` @dependabot is not able to verify dependencies

> Dependabot can't resolve your Go dependency files Dependabot failed to update your dependencies because there was an error resolving your Go dependency files.
>
> Dependabot encountered the following error:
>
> verifying github.com/go-yaml/yaml/v2@v2.4.0/go.mod: checksum mismatch

janisz:
Why this was closed?

PR #540, opened 2019-11-05
chore (v2): update travis config to use latest go versions

chenrui333:
relates to #510 

build before the change, https://travis-ci.org/go-yaml/yaml/builds/607744907

build after the change, https://travis-ci.org/go-yaml/yaml/builds/607744907


niemeyer:
I haven't checked, but I sort of assumed that this was the default behavior of Travis. In either case, this certainly won't hurt.

Thanks for the change.

PR #541, opened 2019-11-05
chore (v3): update travis config to use latest go versions

chenrui333:
relates to #526 

clone of #540


laverya:
Ah, so it uses `1.13.4` instead of `1.13.0`
Makes sense

PR #543, opened 2019-11-07
Port stale simple_keys fix to v2

cjcullen:
Port @niemeyer's https://github.com/go-yaml/yaml/commit/e228e37189d388c1bff077892b97197221e505cb patch back to v2 to fix the issues highlighted in https://github.com/go-yaml/yaml/pull/537.

This PR includes the tests from https://github.com/go-yaml/yaml/pull/537 (as valid YAML this time!)

PR #544, opened 2019-11-08
Stop checking uniqueKeys once we've found 10 errors

cjcullen:
This includes the benchmark test cases from https://github.com/go-yaml/yaml/pull/543.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -746,12 +746,16 @@ func (d *decoder) mapping(n *Node, out reflect.Value) (good bool) {
 	l := len(n.Content)
 	if d.uniqueKeys {
 		nerrs := len(d.terrors)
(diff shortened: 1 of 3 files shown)


cjcullen:
It's kinda overkill to "benchmark" this change, but:
```
$ benchcmp old.txt new.txt
benchmark                    old ns/op      new ns/op     delta
BenchmarkDuplicateKeys-6     2370532130     6586106       -99.72%

benchmark                    old allocs     new allocs     delta
BenchmarkDuplicateKeys-6     2002187        8139           -99.59%

benchmark                    old bytes     new bytes     delta
BenchmarkDuplicateKeys-6     171812904     1012729       -99.41%
```

Issue #551, opened 2019-11-20
Consistent marshaling of generated code such as protobuf would be helpful


ebilling:
I am using protobufs extensively, but protobuf code generation in go does not specify how to handle YAML marshaling.  In this case, I would like to pass the JSON tag to the YAML parser to ensure that the name matches the proto definition itself to ensure consistency across various platforms.

ex.) 

```
type InstanceGroup struct {
	Name              string   'protobuf:"bytes,1,opt,name=name" json:"name,omitempty"'
	NumberOfInstances int32    'protobuf:"varint,2,opt,name=numberOfInstances" json:"numberOfInstances,omitempty"`
}
```

Ideally, I would be able to feed alternative tags to use for marshaling.

bpeake-illuscio:
+1 Would love to see this for my own use case

niemeyer:
I can understand the desire, but JSON is not YAML, and the json package is not the yaml package either. There are several things that one needs to do to make such a parser work correctly and as expected, and I'm not keen on having to forever maintain all of the json package behavior inside the yaml package, and all the awkward edge cases that this will create.

With that said, if you want the behavior of the json package as a YAML file, it's trivial to use the json package to create a generic map of the data, and then ask the yaml package to marshal it out for you.

PR #552, opened 2019-11-20
Allow alternative tags to direct YAML parsing

ebilling:
To deal with issue #551.  By allowing the user to specify other tags to use if the yaml tag is not specified, the user gains more control over yaml output.


niemeyer:
Please see the open issues related to supporting the json tag in yaml.

PR #555, opened 2019-11-26
Optimize cases with long potential simple_keys

cjcullen:
When we build up the simple_keys stack, we count on the (formerly named) staleness check to catch errors where a simple key is required but would be > 1024 chars or span lines. The previous simplification that searches the stack from the top can go 1024 keys deep before finding a "stale" key and stopping. I added a test that shows that this consumes ~3s per 1MB of document size.

I split that staleness check back out into a separate loop so that we don't unnecessarily re-process a bunch of keys just to get down to the ones that might have gone stale.
```
$ benchcmp old.txt new.txt
benchmark                                old ns/op      new ns/op     delta
Benchmark1000KB100Aliases-6              881167484      872120600     -1.03%
Benchmark1000KBDeeplyNestedSlices-6      48761251       5274819       -89.18%
Benchmark1000KBDeeplyNestedMaps-6        50438114       5292240       -89.51%
Benchmark1000KBDeeplyNestedIndents-6     4385726        4280545       -2.40%
Benchmark1000KB1000IndentLines-6         435702849      432047937     -0.84%
Benchmark1KBMaps-6                       574312         588420        +2.46%
Benchmark10KBMaps-6                      5727272        5895964       +2.95%
Benchmark100KBMaps-6                     52126341       52920733      +1.52%
Benchmark1000KBMaps-6                    484546095      474623653     -2.05%
BenchmarkDeepSlice-6                     455902294      397041567     -12.91%
BenchmarkDeepFlow-6                      399613786      407783513     +2.04%
Benchmark1000KBMaxDepthNested-6          2700904018     447527143     -83.43%

benchmark                                old allocs     new allocs     delta
Benchmark1000KB100Aliases-6              5832425        5832415        -0.00%
Benchmark1000KBDeeplyNestedSlices-6      9066           10081          +11.20%
Benchmark1000KBDeeplyNestedMaps-6        9071           10087          +11.20%
Benchmark1000KBDeeplyNestedIndents-6     10082          10082          +0.00%
Benchmark1000KB1000IndentLines-6         4093516        4093516        +0.00%
Benchmark1KBMaps-6                       3126           3126           +0.00%
Benchmark10KBMaps-6                      30780          30780          +0.00%
Benchmark100KBMaps-6                     307269         307269         +0.00%
Benchmark1000KBMaps-6                    3072079        3072079        +0.00%
BenchmarkDeepSlice-6                     2048121        2048112        -0.00%
BenchmarkDeepFlow-6                      1978104        1978103        -0.00%
Benchmark1000KBMaxDepthNested-6          4079998        4079989        -0.00%

benchmark                                old bytes     new bytes     delta
Benchmark1000KB100Aliases-6              393118344     392792852     -0.08%
Benchmark1000KBDeeplyNestedSlices-6      4689465       4576336       -2.41%
Benchmark1000KBDeeplyNestedMaps-6        4689831       4577450       -2.40%
Benchmark1000KBDeeplyNestedIndents-6     2969924       2969913       -0.00%
Benchmark1000KB1000IndentLines-6         143718512     143718512     +0.00%
Benchmark1KBMaps-6                       218984        218984        +0.00%
Benchmark10KBMaps-6                      2178155       2178154       -0.00%
Benchmark100KBMaps-6                     22002796      22002807      +0.00%
Benchmark1000KBMaps-6                    220560496     220560496     +0.00%
BenchmarkDeepSlice-6                     120114888     119789000     -0.27%
BenchmarkDeepFlow-6                      115058032     115056832     -0.00%
Benchmark1000KBMaxDepthNested-6          146582888     146257064     -0.22%
```

### DIFF
--- a/scannerc.go
+++ b/scannerc.go
@@ -626,36 +626,17 @@ func trace(args ...interface{}) func() {
 func yaml_parser_fetch_more_tokens(parser *yaml_parser_t) bool {
 	// While we need more tokens to fetch, do it.
 	for {
-		// Check if we really need to fetch more tokens.
-		need_more_tokens := false
-
-		if parser.tokens_head == len(parser.tokens) {
-			// Queue is empty.
-			need_more_tokens = true
-		} else {
-			// Check for any now stale keys in a required position.
-			if !yaml_parser_stale_simple_keys(parser) {
+		if parser.tokens_head != len(parser.tokens) {
+			// If queue is non-empty, check if any potential simple key may
+			// occupy the head position.
+			head_tok_idx, ok := parser.simple_keys_by_tok[parser.tokens_parsed]
+			if !ok {
+				break
+			} else if valid, ok := yaml_simple_key_is_valid(parser, &parser.simple_keys[head_tok_idx]); !ok {
 				return false
+			} else if !valid {
+				break
 			}
-			// Check if any potential simple key may occupy the head position.
-			for i := len(parser.simple_keys) - 1; i >= 0; i-- {
-				simple_key := &parser.simple_keys[i]
-				if !simple_key.possible {
-					break
-				}
-				if simple_key.token_number < parser.tokens_parsed {
-					break
-				}
-				if simple_key.token_number == parser.tokens_parsed {
-					need_more_tokens = true
-					break
-				}
-			}
-		}
-
-		// We are finished.
-		if !need_more_tokens {
-			break
 		}
 		// Fetch the next token.
 		if !yaml_parser_fetch_next_token(parser) {
@@ -864,20 +845,6 @@ func yaml_simple_key_is_valid(parser *yaml_parser_t, simple_key *yaml_simple_key
 	return true, true
 }
 
-func yaml_parser_stale_simple_keys(parser *yaml_parser_t) bool {
-	// fmt.Println(parser.simple_keys_min_possible_index, ":", len(parser.simple_keys))
-	for i := parser.simple_keys_min_possible_index; i < len(parser.simple_keys); i++ {
-		simple_key := &parser.simple_keys[i]
-		if valid, ok := yaml_simple_key_is_valid(parser, simple_key); !ok {
-			return false
-		} else if !valid {
-			parser.simple_keys_min_possible_index = i + 1
-		}
-
-	}
-	return true
-}
-
 // Check if a simple key may start at the current position and add it if
 // needed.
 func yaml_parser_save_simple_key(parser *yaml_parser_t) bool {
@@ -902,6 +869,7 @@ func yaml_parser_save_simple_key(parser *yaml_parser_t) bool {
 			return false
 		}
 		parser.simple_keys[len(parser.simple_keys)-1] = simple_key
+		parser.simple_keys_by_tok[simple_key.token_number] = len(parser.simple_keys) - 1
 	}
 	return true
 }
@@ -918,9 +886,7 @@ func yaml_parser_remove_simple_key(parser *yaml_parser_t) bool {
 		}
 		// Remove the key from the stack.
 		parser.simple_keys[i].possible = false
-		if parser.simple_keys_min_possible_index > i {
-			parser.simple_keys_min_possible_index = i
-		}
+		delete(parser.simple_keys_by_tok, parser.simple_keys[i].token_number)
 	}
 	return true
 }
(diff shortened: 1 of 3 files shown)


scannerc.go
```diff
@@ -859,6 +864,20 @@ func yaml_simple_key_is_valid(parser *yaml_parser_t, simple_key *yaml_simple_key
 	return true, true
 }
 
+func yaml_parser_stale_simple_keys(parser *yaml_parser_t) bool {
+	// fmt.Println(parser.simple_keys_min_possible_index, ":", len(parser.simple_keys))
```
liggitt: drop debugging?

scannerc.go
```diff
@@ -633,15 +633,20 @@ func yaml_parser_fetch_more_tokens(parser *yaml_parser_t) bool {
...
+			// Check for any now stale keys in a required position.
+			if !yaml_parser_stale_simple_keys(parser) {
+				return false
+			}
 			// Check if any potential simple key may occupy the head position.
 			for i := len(parser.simple_keys) - 1; i >= 0; i-- {
 				simple_key := &parser.simple_keys[i]
+				if !simple_key.possible {
+					break
+				}
```
niemeyer: Again, it's possible I misunderstand the algorithm since I didn't put the original logic in place here, but this doesn't look right to me. That's stopping the search if the most recently scanned depth level was marked as not a possible simple key, but that's not the actual question being asked by the logic. The question is: can the head token -- that is, the token that is about to be consumed by the parser -- be a simple key still, given everything else we parsed so far? We can't tell if the head token is possible or not without finding it, which is done below.

This will certainly speed things up because it's not trying to find the head anymore, but it doesn't seem right.

Am I missing something?

scannerc.go
```diff
@@ -859,6 +864,20 @@ func yaml_simple_key_is_valid(parser *yaml_parser_t, simple_key *yaml_simple_key
 	return true, true
 }
 
+func yaml_parser_stale_simple_keys(parser *yaml_parser_t) bool {
+	// fmt.Println(parser.simple_keys_min_possible_index, ":", len(parser.simple_keys))
+	for i := parser.simple_keys_min_possible_index; i < len(parser.simple_keys); i++ {
```
niemeyer: I'd like to avoid adding this linear search back. I understand why it existed and why it's being suggested again, but in principle another linear search is worse than fixing the same issue without having it. But let's discuss the earlier issue first and then I'll think about this case some more.

niemeyer (changes requested):
Thanks for keeping it up.

Here are some initial comments on the issue:

scannerc.go
cjcullen: I think you're right. I'm not sure what the negative test case would be, but I agree that the spirit of this piece of logic seems to not be maintained.

From what I can tell, this function's intent is to fetch enough tokens to determine whether the next value fetched for the parser is going to be an implicit simple-key (e.g. `[[[]]]: foo`). If can't be a simple key, and that was required, fail. If it can't be a simple key but we didn't require one, then let the parser consume the non-simple-key value.

If we want to keep the check for "can a simple key occupy the head position" as a simple scan, I think we always end up w/ potentially 1024 keys to scan, before we either decide we need more tokens or that we are ready to return w/ a token available (or if we scan from the bottom, up to 10,000 keys to scan).

If the only key we actually care about is the one that potentially starts with the token at the parser head, then we only really need to check that. So now I am back to my other proposal of maintaining a map of possible simple_keys :).

scannerc.go
cjcullen: Ack. I think this is unnecessary if we can always count on checking the token at the head of the parser for staleness.

scannerc.go
cjcullen: Ack.

cjcullen:
Thanks again for the first pass. Are you up for considering the addition of the single_key index? I'd like to have something to include in the next set of Kubernetes patch releases around the first week of January.

scannerc.go
niemeyer: > If the only key we actually care about is the one that potentially starts with the token at the parser head, then we only really need to check that. So now I am back to my other proposal of maintaining a map of possible simple_keys :).

Yes, your map proposal was never considered incorrect. It's a reasonable way to handle a very complex case. The reason why I've suggested resisting it is because as far as I can see this sort of pattern will only be found in unusual documents, typically abusive. Real documents will pretty much never get even remotely close to this being an issue.

That said, if that's what it takes, so be it, and the change on top of the new logic also looks simpler compared to the previous proposal we looked into.

I'll try to merge this today or tomorrow at a window that allows me to pay attention to any issues.

Abuchtela:
Update gopkg.in/yaml.v2

PR #557, opened 2019-11-28
Add decoder options type

vsapronov:
Here's the problem description:

1. I need to maintain order in my maps, so I use something like `type Items []NamedItem` to represent a map through array.
2. Because `Items` should be represented as map in yaml I need to implement custom `(a *Items) UnmarshalYAML(node *yaml.Node)`
3. In my custom unmarshal method I just get key nodes and value nodes ordered from `node.Content`, then I decode each value node with automatic fields mapping into corresponding structure with `node.Decode`
4. Unfortunately I can't ask `node.Decode` to enable `KnownFields` mode - there's no any way to do it.

Here are changes made in this PR:

1. There was unused parameter `srtict` in `unmarshal` method here: https://github.com/go-yaml/yaml/compare/v3...vsapronov:v3?expand=1#diff-e51f89b39d976550831afb510a926be0L156. I have removed it.
2. `DecodeOptions` was introduced as a struct representing decoding options. It includes `knownFields` and also `uniqueKeys` settings. It's very easy to add any other decode setting through `DecodeOptions` if needed. `DecodeOptions` supports "builder" style creation and initialization to make instantiation easier: `NewDecodeOptions().KnownFields(true).UniqueKeys(true)`. 
3. `DecodeOptions` is set to internal decoder in all places where it's passed (as opposed to `strict` parameters from up above). To maintain backward compatibility alternative methods are introduced: `yaml.UnmarshalWith`,  `yaml.NewDecoderWith` and `node.DecodeWith`: https://github.com/go-yaml/yaml/pull/557/files#diff-e51f89b39d976550831afb510a926be0R92
https://github.com/go-yaml/yaml/pull/557/files#diff-e51f89b39d976550831afb510a926be0R117
https://github.com/go-yaml/yaml/pull/557/files#diff-e51f89b39d976550831afb510a926be0R158.
Internal `newDecoder` is using default `DecodeOptions` if passed one is `nil`.


vsapronov:
@niemeyer Please let me know your opinion on this proposed change and the issue.

vsapronov:
@niemeyer Any chance you could take a look at this?

decode.go
```diff
@@ -328,7 +349,7 @@ func newDecoder() *decoder {
 	d := &decoder{
 		stringMapType:  stringMapType,
 		generalMapType: generalMapType,
-		uniqueKeys:     true,
```
niemeyer, line 339: This is enforced by the spec. I'm not sure we want to allow tuning it anymore. There's just no good reason to accept broken documents.

decode.go
```diff
@@ -299,6 +299,28 @@ func (p *parser) mapping() *Node {
 // ----------------------------------------------------------------------------
 // Decoder, unmarshals a node into a provided value.
 
+// DecodeConfig represents configuration for decoding.
+// Config could be initialized in "builder" style for convenience:
+// NewDecodeConfig().KnownFields(true).UniqueFields(true)
+type DecodeConfig struct {
```
niemeyer: Options might be better than Config, as the former is a real word.

decode.go
```diff
@@ -299,6 +299,28 @@ func (p *parser) mapping() *Node {
 // ----------------------------------------------------------------------------
 // Decoder, unmarshals a node into a provided value.
 
+// DecodeConfig represents configuration for decoding.
+// Config could be initialized in "builder" style for convenience:
+// NewDecodeConfig().KnownFields(true).UniqueFields(true)
```
niemeyer: One of the reasons why I've used the method style in the encoder and decoder is that it allows us to protect against modification of the configuration out of turn. It also means we can use an initializer such as NewEncoder/NewDecoder, which prevents us having to test for initialization everywhere else.

That said, the Options type will by definition be a type prepared ahead of the creation. It seems like it would benefit from using a structure with public fields instead. It should have just one private field to force a non-positional instantiation.

yaml.go
```diff
@@ -141,6 +152,20 @@ func (dec *Decoder) Decode(v interface{}) (err error) {
 // conversion of YAML into a Go value.
 func (n *Node) Decode(v interface{}) (err error) {
 	d := newDecoder()
+	return n.decode(d, v)
+}
+
+// DecodeWithConfig decodes the node and stores its data into the value pointed to by v.
+// It uses passed DecodeConfig to setup internal decoder.
+func (n *Node) DecodeWithConfig(v interface{}, config *DecodeConfig) (err error) {
```
niemeyer: If we invert the order of fields, we might name it as just _DecodeWith_, which would make it read nicely as in:
```
err := node.DecodeWith(options, &value)
```

We should probably also have other functions and methods taking options, but it's okay to start with just these.

niemeyer (changes requested):
Some initial brainstorm follows.

Have you signed the contributor agreement yet?

https://ubuntu.com/legal/contributors

Thanks!

decode.go
vsapronov: Renamed to DecodeOptions

yaml.go
vsapronov: This is done: `yaml.UnmarshalWith`, `yaml.NewDecoderWith` and `node.DecodeWith` are added.

decode.go
vsapronov: Do you think we just need to remove control over it? In proposed design it's just a matter or removal `UniqueKeys` method from `DecodeOptions` and it will be not changeable. Though it's strange that it's used in code 3 times if this feature is not needed - may be it would be worth to remove all corresponding `if d.options.uniqueKeys` statements.

decode.go
vsapronov: Not fully sure that I follow this. I added `DecodeOptions` argument to the internal `newDecoder` function. Is that what you meant above?

vsapronov:
@niemeyer Made changes as requested. Also my contributor agreement is signed for my Github account. Could you please take a look at current version of these changes.

gtrewitt:
+1
I would really like to see this get implemented
**Note:** Although Adding Node.KnownFields() is a fine solution, it would be nice if Node could inherit the setting from the Decoder.  But anything is better than nothing.

vsapronov:
@niemeyer could you please review this again

Emyrk:
@vsapronov thanks for taking this on. I am thinking about forking or choosing another decode library because I cannot control the decode options when decoding from a `Node`. This PR allows me to continue using this package

h17liner:
Hi!, Any updates?

vsapronov:
Heads up everyone who is using my fork from this PR: I had to move it to the Github organization where I really need it: specgen-io. Therefore now if you want to use it the full module name has changed to `gopkg.in/specgen-io/yaml.v3` (NO `vsapronov` there anymore). All best.
I really hope that one day Gustavo @niemeyer would approve this PR.

vsapronov:
E6vXp1EXIAIlDKJ

niemeyer:
That's a nice one. :-)

I'm still here, didn't forget.. sorry that this is taking more time than I hoped. It also bothers me to not be able to give you the due attention, but I realize that this doesn't make it much better on your end.

Issue #512, opened 2019-10-01
Allow (un)marshal with parameters that apply to all fields


jpap:
It is helpful to apply some flags to all fields being (un)marshaled.  For example, to be able to say that all fields should be included if it is not the zero value, without having to replicate the `yaml:",omitempty"` tag everywhere.

The [asn.1 package](https://golang.org/pkg/encoding/asn1/) does this nicely by having overloaded (un)marshal functions that you can provide these global flags as a comma-separated string, in the same form as the usual field tag. That approach may be appropriate here also.

Issue #500, opened 2019-09-06
Allow custom field name marshalling function


tliron:
This is a feature request.

Currently (v3), reflection-based marshalling calls `ToLower` on all field names. The exact call is done [here](https://github.com/go-yaml/yaml/blob/v3/yaml.go#L590). One can override the value by setting the yaml tag, but this must be done on every field, everywhere, which could be impossible to do for 3rd-party libraries.

Would it be possible to support a `SetFieldNameTranslator` function for the marshaller, or a similar mechanism? This would allow for consistent translation, for example to camel case, to underscore separation of words, etc.

msteffen:
Would https://github.com/go-yaml/yaml/issues/424 meet the needs of your use-case? I ask because I'm interested in that feature as well.

tliron:
@msteffen Unfortunately no, because #424 would require tagging all fields. I'm specifically looking to change the *generic* "get field name" function that is used on un-tagged fields. Although, it does seem like the discussion there might benefit from my suggestion here. :)

Issue #123, opened 2015-09-08
Feature request: case-insensitive keys when Unmarshalling ?


moul:
So we can load files like these:

``` yaml
hosts:
  toto:
    hostname: 1.2.3.4
    port: 42
```

As well as

``` yaml
Hosts:
  toto:
    HostName: 1.2.3.4
    Port: 42
```

Using

``` go
type File struct {
  Hosts: []Host `yaml:Hosts`,
}

type Host struct {
  HostName: string `yaml:HostName`
  Port: int `yaml:Port`
}
```

dhiltonp:
For this, I'm running ToLower on the input data; considering that case sensitivity is part of the yaml spec, I think it's a better solution than changing the library (although it may be less convenient).

jpap:
@dhiltonp, would love if you have a little time to review my PR so we can resolve this 3+ year old issue.  I've also run into the same problem, where I'd like to ignore case-sensitivity on unmarshal.

moul:
Additional note, it's important to only ignore case-sensitivity for the key; the value should not be lowered

As a workaround I made this: https://github.com/moul/flexyaml

I would love to depreciate this repo in favor of an official solution :)

jpap:
@moul, my PR only considers case insensitivity for the key; the value is kept as-is.

Looking at the mounting PRs backed up over a year, I'm wondering whether this project is still being maintained.  I'm going to be working from a fork, unfortunately.  In your case, you might do the same; feel free to include my PR if that suits your needs.

frickenate:
What the actual hell. We all understand the limitation of the first character having to be uppercase due to export requirements... but yaml is case-sensitive. Where the **_hell_** does "unmarshalled using the field name lowercased as the default key" come from?!

You're making me tag every.... single... fkn field that has an uppercase character in it. Clearly my `InstallVersion` field should be accepting the yaml field for `installVersion`... not `installversion`. For fuck's sake. Almost every... damn... field... has to be remapped manually. Not because my fields are strange, but because you arbitrarily decided on some absurd and broken "lowercase all the things".

Fix this in a v3 please. Backwards-incompatible as hell, but you'll be restoring expected behaviour. We're not even asking for complete case-insensitivity. Just that beyond that first unfortunate letter due to exported fields, you don't touch the rest of the string.

[...]

PR #563, opened 2019-12-14
add struct info options

cizmazia:
I have drafted injection of struct info options to make them configurable. Are you okay with this approach?

I have tested marshaling/unmarshaling as follows:
```go
func MustParseConfig(s string) *Config { 
    var cfg Config                                                                                 
    err := yaml.Unmarshal([]byte(s), &cfg, yaml.DecoderWithFieldNameMarshaler(FieldNameMarshaler))
    if err != nil {
        panic(fmt.Errorf("invalid config: %v", err))                                         
    }                                                                                               
    return &cfg                                                                                     
}                                                                                                   
                                                                                                    
func (c *Config) String() string {                                                                  
    b, _ := yaml.Marshal(c, yaml.EncoderWithFieldNameMarshaler(FieldNameMarshaler))                        
    return string(b)                                                                                
}                                                                                                   
                                                                                                    
func FieldNameMarshaler(f reflect.StructField) string {                                             
    return f.Name
}
```

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -246,10 +246,36 @@ var (
 	ptrTimeType    = reflect.TypeOf(&time.Time{})
 )
 
-func newDecoder(strict bool) *decoder {
-	d := &decoder{mapType: defaultMapType, strict: strict, structParser: DefaultStructParser}
+func newDecoder(opt ...DecoderOption) *decoder {
+	d := &decoder{mapType: defaultMapType, structParser: DefaultStructParser}
+	for _, o := range opt {
+		o(d)
+	}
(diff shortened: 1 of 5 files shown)


cizmazia:
Related to #500, #512, #424, #123

niemeyer:
Yes, something along those lines will probably go in eventually given the interest we have. There are several small issues in the draft, though, which I won't bother you with as it's a draft. We should address the decoder/encoder options first, which is already on going in a different PR, so we can have a nice place to extend these routines.

PR #564, opened 2019-12-15
Feature (v3): compatible with only json-tag

junhwong:
In some scenarios, you cannot manually add the `yaml` tag, such as grpc's automatic code generation. Such as:
```go
// Code generated by protoc-gen-go. DO NOT EDIT.
//...
type Metadata struct {
	Name                 string            `protobuf:"bytes,1,opt,name=name,proto3" json:"name,omitempty"`
        //...
}
```

work flow：yaml(go struct) -> grpc -> server

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,5 @@
-module "gopkg.in/yaml.v3"
+module gopkg.in/yaml.v3
 
-require (
-	"gopkg.in/check.v1" v0.0.0-20161208181325-20d25e280405
-)
+require gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405
+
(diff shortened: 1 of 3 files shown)


laverya:
This would greatly simplify some of my usage. My only worry is that this might be incompatible with how custom json marshallers are implemented, but even then this would still be better than nothing

junhwong:
Yeah. Regarding compatibility issues, this definitely exists.
but as long as the first priority is yaml, it will only match other if yaml is not defined, such as json.

niemeyer:
Thanks for the effort, but please see discussions in existing issues. I'm not ready to support that at this time.

PR #567, opened 2020-01-05
Fix typo in Decoder docstring.

ktnyt:
I just realized this small typo while I was skimming through the documentation. I could've opened an issue but thought it'd be a hassle.


niemeyer:
Good catch, thanks.

PR #574, opened 2020-02-05
Encode strings with YAML 1.1 boolean values with quotes

dsharp-pivotal:
To allow better compatibility with YAML 1.1 parsers, when outputting a
string that is equal to one of the YAML 1.1 boolean values ("yes", "no",
"on", "off", etc), enclose the string in quotes. The string will then
parse correctly by either YAML 1.1 or YAML 1.2 parsers.

This change does not affect decoding: Bare `yes` and `no` will still be
decoded as strings.

See: #214

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -298,6 +298,23 @@ func isBase60Float(s string) (result bool) {
 // is bogus. In practice parsers do not enforce the "\.[0-9_]*" suffix.
 var base60float = regexp.MustCompile(`^[-+]?[0-9][0-9_]*(?::[0-5]?[0-9])+(?:\.[0-9_]*)?$`)
 
+// isYaml11Bool returns whether s is a boolean value as defined in YAML 1.1.
+//
+// The extra boolean values in YAML 1.1 are a terrible idea and is unsupported
+// in YAML 1.2 and by this package, but these should be marshalled quoted for
+// the time being for compatibility with other parsers.
+func isYaml11Bool(s string) (result bool) {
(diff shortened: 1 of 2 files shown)


laverya:
Is this functionally different than https://github.com/go-yaml/yaml/pull/490?

(I don't care which fix gets merged. I just want a fix to get merged)

dsharp-pivotal:
@laverya Oh, I didn't see your PR.

They look pretty similar, but I see one functional difference: This PR also forces quotes when a node is explicitly tagged as `!!str` (the diff in `encoder.node()`). My desire for that specifically is described in mikefarah/yq#341

laverya:
They both make the same (functional) change to encoder.node though? 

```
		canUsePlain = rtag == strTag && !isBase60Float(s) && !isYaml11Bool(s)
```
vs
```
		canUsePlain = rtag == strTag && !isBase60Float(s)

		// Check to see if this is a v1.1 yaml bool string
		// if it is, require quotes to preserve backwards compatibility
		_, is11BoolString := boolMap[s]
		canUsePlain = !is11BoolString && canUsePlain
```

Not that it matters - as I said before, as long as one gets merged I'm happy.

dsharp-pivotal:
@laverya Yeah, in terms of those changes, looks equivalent to me.

@niemeyer Can you take a look?

diamondburned:
https://github.com/go-yaml/yaml/pull/583 actually left out the change in `node()` that this PR includes.

Issue #582, opened 2020-03-12
v3  marshalling of  "boolean" strings breaks YAML 1.2 interoperability.


tcolgate:
In v2:
```go
package main

import (
        "fmt"
        "log"

        yaml "gopkg.in/yaml.v2"
        //yaml "gopkg.in/yaml.v3"
)

func main() {
        data := struct {
                ID string `yaml:"id"`
        }{ID: "off"}
        bs, err := yaml.Marshal(data)
        if err != nil {
                log.Fatalf("err: %v", err)
        }
        fmt.Println(string(bs))
}
```

this prints:
```
id: "off"
```

in v3:
```
id: off
```

The field is a string, and should marshal, and re-unmarshal as one.

tcolgate:
I'm pretty confused now, I can't recreate this in the tests. I've checked the commit I'm using, it's a6ecf24a6d71 , which is the same as the checkout I'm running the tests on.

tcolgate:
okay, the test I added was bad. it works if the string is "true", it doesn't work if the string is "off"

tcolgate:
okay, it has just occurred to me as to what is happening here. I saw a comment around the yaml spec change for these values, and you are now only handling true/false with the quote enforced marshaling.
This means the output effectively needs to be read with a reader that also adheres to the spec, and cleanly.
I'm not really sure how/if to fix this. Ideally we'd marshal these out as quoted, because it fixes my problem, but the world doesn't revolve around me.

tcolgate:
So, in a sort of begging/pleading sense, I would suggest that marshaling the old 1.2 "boolean" strings as quoted strings would not, strictly speaking, be a break from the spec maybe?

niemeyer:
Yes, you're right and the change makes sense. We should quote strings that match 1.1 booleans.

PR #583, opened 2020-03-13
Marshal YAML 1.1 bool-like to strings

tcolgate:
I'm not 100% sure if this is the right place to do this, it works, and
is somewhat in the spirit of decode.go.  I've put an explanation of the
reasoning in the commit, not sure if you'd rather have something in the README.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -298,6 +298,21 @@ func isBase60Float(s string) (result bool) {
 // is bogus. In practice parsers do not enforce the "\.[0-9_]*" suffix.
 var base60float = regexp.MustCompile(`^[-+]?[0-9][0-9_]*(?::[0-5]?[0-9])+(?:\.[0-9_]*)?$`)
 
+// isOldBool returns whether s is bool notation as defined in YAML 1.1.
+//
+// We continue to force strings that YAML 1.1 would interpret as booleans to be
+// rendered as quotes strings so that the marshalled output valid for YAML 1.1
+// parsing.
+func isOldBool(s string) (result bool) {
+	switch s {
+	case "y", "Y", "yes", "Yes", "YES", "on", "On", "ON",
+		"n", "N", "no", "No", "NO", "off", "Off", "OFF":
+		return true
+	default:
+		return false
+	}
+}
+
 func (e *encoder) stringv(tag string, in reflect.Value) {
 	var style yaml_scalar_style_t
 	s := in.String()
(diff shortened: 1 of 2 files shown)


tcolgate:
Fixes #582

encode.go
```diff
@@ -320,6 +320,15 @@ func (e *encoder) stringv(tag string, in reflect.Value) {
 		// there's no need to quote it.
 		rtag, _ := resolve("", s)
 		canUsePlain = rtag == strTag && !isBase60Float(s)
+
+		// Maintain some backward compatibility with YAML 1.2
+		if rtag == strTag {
```
niemeyer: This would be slightly nicer if we had it inside something like _isOldBool_ and put the test next to the isBase60Float above, as that method is the exactly equivalent except for base 60 floats (!!!, and you thought the on/off situation was bad).

Checking for isBase60Float usage should find us the exact places we need to care about. Well, either that or we have two bugs instead of one.. ;-)

niemeyer (changes requested):

encode.go
tcolgate: Okay, have moved the check out and added the test to the above line. it does read more clearly.

niemeyer:
Thanks, glad to merge this.

Have you signed the contributor agreement before?  It's trivial to sign online, and it's a non-draconian licensing with you retaining ownership:

https://ubuntu.com/legal/contributors

tcolgate:
I think I just signed the form, I put you down as "Canonical Project Manager or contact"

niemeyer:
Thanks!

niemeyer:
For the record, please note that those are YAML 1.1 strings, vs. the new 1.2 style that is more strictly implemented in go-yaml v3 (not 1.2 and 1.3).

Thanks again for your help.

tcolgate:
Right, thanks. I had it my head the "fix" to the spec was in 1.3, I think I saw that on the issue that pointed out they've been removed.  I think I fixed the comments and the PR title, but not the commit message.
Thanks very much for go-yaml v3, exposing the AST made a port of a python tool to Go we have possible.

laverya:
IMO this removes the need for https://github.com/go-yaml/yaml/pull/574 and https://github.com/go-yaml/yaml/pull/490, too - happy to see one of these finally get merged!

laverya:
(And fixes issues #489  and #214)

PR #585, opened 2020-03-22
Add go1.14 into travis config

chenrui333:
The latest go1.14 is v1.14.1, it would be good to add it as cross-check as part of travis build.


niemeyer:
Thanks again.

chenrui333:
Thanks for the merge!

PR #589, opened 2020-03-27
Support marshalling yaml.Node values

### LINKED ISSUE #588: Marshaling of yaml.Node value not supported, *yaml.Node pointer is supported
(v3 branch)

encode.go has support for marshaling `*yaml.Node` in structs, but not `yaml.Node` values:

https://github.com/go-yaml/yaml/blob/v3/encode.go#L119

This seems to be an oversight, as for other data types like `time.Time`, both the pointer and the value version are supported.

I can provide a pull request for this problem which adds support and test cases for the value and the pointer version of `yaml.Node`, if this would be accepted.

worldturner:
Fixes #588

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -116,7 +116,7 @@ func (e *encoder) marshal(tag string, in reflect.Value) {
 	}
 	iface := in.Interface()
 	switch value := iface.(type) {
-	case *Node:
+	case *Node, Node:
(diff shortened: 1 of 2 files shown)


niemeyer:
Thank you. This was fixed and tested with a81ed60d.

PR #593, opened 2020-04-03
Extend Node with start index for all types and end markers for scalars

atomatt:
Copies additional marker fields from the parser's events to the relevant Node
instance to expose additional source locations in the public API.

Index is populated form the start event for all node types. Unlike Line/Column,
Index is a zero-based (rune) offset from the start of the YAML stream.

LineEnd, ColumnEnd and IndexEnd are populated from the end event. The fields
have the same semantics as the equivalent start marker fields.

For now, *End locations are only set/meaningful for scalars because end marker
info at the parser level seems spurious in some situations. However, the main
use case for exposing end markers is to make in-place edits to YAML source, and
scalars are the most likely targets, so this seems like a reasonable compromise
and therefore a sensible first step.


yaml.go
```diff
@@ -389,6 +389,21 @@ type Node struct {
 	// These fields are not respected when encoding the node.
 	Line   int
 	Column int
+
+	// Index holds the node's start position in the decoded YAML text as a
+	// zero-based offset. This fields is not respected when encoding the node.
+	Index int
```
mkmik, line 395: should we be explicit here that this is a run offset and not a byte offset?

atomatt: Maybe. The doc strings for new fields mostly just follow the style of the existing Line/Column doc string so those would also need to be updated.

yaml.go
```diff
@@ -389,6 +389,21 @@ type Node struct {
...
+	// LineEnd and ColumnEnd hold the node's end position in the decoded YAML
+	// text. These fields are only set for scalars. These fields are not
+	// respected when encoding the node.
+	LineEnd   int
+	ColumnEnd int
+
+	// IndexEnd hold the node's end position in the decoded YAML text as a
+	// zero-based offset. This field is only set for scalars. This field is not
+	// respected when encoding the node.
+	IndexEnd int
```
mkmik, line 406: This appears to include the trailing newline when a string is in literal and folded style.

I.e. consider two equivalent yamls:

```
foo: bar
# end
```

`IndexEnd` of node `/foo` is `8`.  The rune at 8th position is the "\n" following bar

while:

```
foo: |-
  bar
# end
```
`IndexEnd` of node `/foo` is `14`.  The rune at 14th position is the "#" following the newline after bar.

I don't think that's correct. The trailing newline in the source yaml doesn't technically belong to a "scalar" value; instead it's the delimiter that separates the value from the next field.
Multiline fields do allow multiple newlines, but that's because intervening indentation overrides the terminator-like semantics of the newline.
To further confuse the waters, the semantics of a `|` multiline block is to include a trailing newline in the node value, while "|-" does not. I _think_ the correct way to look at this is not that "|-" strips the newline but instead that "|" adds a trailing newline that wouldn't otherwise be there in the value.

My main argument is consistency. No matter what format the node has, the range Index-IndexEnd should fully cover the body of the scalar and removing that rune slice from the input should yield a valid yaml where the field is empty.

---

Go code to reproduce

```go
package main

import (
	"fmt"
	"log"
	"os"
	"text/tabwriter"

	"gopkg.in/yaml.v3"
)

func main() {
	testCases := []string{
		`foo: bar
# end`,
		`foo: |-
  bar
# end
`,
	}

	w := tabwriter.NewWriter(os.Stdout, 0, 0, 1, ' ', 0)

	for i, src := range testCases {
		fmt.Printf("\nParsing test case %d:\n", i)
		var n yaml.Node
		if err := yaml.Unmarshal([]byte(src), &n); err != nil {
			log.Fatal(err)
		}
		v := n.Content[0].Content[1]
		end := v.IndexEnd
		fmt.Fprintf(w, "Value:\t%q\n", v.Value)
		fmt.Fprintf(w, "IndexEnd:\t%d\n", end)
		fmt.Fprintf(w, "src[.IndexEnd]:\t%q\n", src[end])
		w.Flush()
	}
}
```

```
Parsing test case 0:
Value:          "bar"
IndexEnd:       8
src[.IndexEnd]: '\n'

Parsing test case 1:
Value:          "bar"
IndexEnd:       14
src[.IndexEnd]: '#'
```

PR #595, opened 2020-04-04
The configuration value of string type supports the syntax of environment variables.

yankooo:
The configuration value of the string type supports the following syntax :

demo.yaml :
``` yaml
gopath: ${GO||/usr/local/go}
```
if environment variable `${Go}` isn't setting,  return default value `/usr/local/go` .

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -432,6 +432,9 @@ func (d *decoder) scalar(n *node, out reflect.Value) bool {
 		}
 		return true
 	}
+	if str, ok := resolved.(string); ok {
+		resolved = expandValueEnv(str)
+	}
 	if resolvedv := reflect.ValueOf(resolved); out.Type() == resolvedv.Type() {
 		// We've resolved to exactly the type we want, so use that.
 		out.Set(resolvedv)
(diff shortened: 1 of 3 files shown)


laverya:
How do you propose to encode the literal string `${GO||/usr/local/go}`?

I think this would be a better proposal for [yaml itself](https://yaml.org/) than for a specific yaml parser library.

yankooo:
The developer needs to get an environment variable for some configuration, like `service name`、`container IP` etc, when deploying the project to docker or other virtual containers. If you get the environment variable in business code for that special configuration, it is not convenient. Cause you must code many `if-else` in business code for that configuration.

So, I propose to encode the literal string ${GO||/usr/local/go}, to convenient deploy project. the developer just needs to point out in the config file of `YAML`. Like I need to get environment real value for ServiceName, just code like `${SERVICE_NAME||web-server}` in my config file of `YAML`.

laverya:
I understand your goal. I'm saying that this project implements the YAML spec, and that is not in the YAML spec.

yankooo:
I understand, thanks.

PR #600, opened 2020-04-14
support convert from yaml.Node

Yohox:
marshal not convert from yaml.Node just convert *yaml.Node
so encode.go 122 add :
	case Node:
		e.nodev(in.Addr())
		return

just work.


niemeyer:
Thanks. This was fixed and tested in a81ed60d.

PR #608, opened 2020-05-11
Fix Marshal documentation on omitempty and IsZero

### LINKED ISSUE #607: Marshal documentation on IsZero in omitempty is wrong
[This line of v2](https://github.com/go-yaml/yaml/blob/0b1645d91e851e735d3e23330303ce81f70adbe3/yaml.go#L178) and [this line of v3](https://github.com/go-yaml/yaml/blob/2ff61e1afc866138abf1a8adf3cc89721090ac31/yaml.go#L197) reads as

     omitempty    Only include the field if it's not set to the zero
                  value for the type or to empty slices or maps.
                  Zero valued structs will be omitted if all their public
                  fields are zero, unless they implement an IsZero
                  method (see the IsZeroer interface type), in which
                  case the field will be included if that method returns true.

**`included`** in the last line should be **`excluded`**.

The current documentation contradicts the source code at [v2](https://github.com/go-yaml/yaml/blob/0b1645d91e851e735d3e23330303ce81f70adbe3/encode.go#L221) and [v3](https://github.com/go-yaml/yaml/blob/2ff61e1afc866138abf1a8adf3cc89721090ac31/encode.go#L225)

afriza:
Fixes #607

### DIFF
--- a/yaml.go
+++ b/yaml.go
@@ -175,7 +175,7 @@ func unmarshal(in []byte, out interface{}, strict bool) (err error) {
 //                  Zero valued structs will be omitted if all their public
 //                  fields are zero, unless they implement an IsZero
 //                  method (see the IsZeroer interface type), in which
-//                  case the field will be included if that method returns true.
+//                  case the field will be excluded if IsZero returns true.
 //
 //     flow         Marshal using a flow style (useful for structs,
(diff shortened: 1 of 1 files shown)


niemeyer:
Thanks again.

PR #622, opened 2020-06-01
Improve sorting consistency of map keys

dominikgrygiel:
This PR would improve consistency when marshaling maps. The problem occurs when comparing letters to other characters, ex. `"a10"` and `"a1a"`. Currently, the result would be non-deterministic because of https://github.com/go-yaml/yaml/blob/0b1645d91e851e735d3e23330303ce81f70adbe3/sorter.go#L49-L51

if `"a10"` happens to be assigned to `ar` and `"a1a"` to `br`, that code would return `true`, while if the assignment was the other way around, it would return `false`.

### DIFF
--- a/encode_test.go
+++ b/encode_test.go
@@ -580,8 +580,8 @@ func (s *S) TestSortedOutput(c *C) {
 		"a/11",
 		"a/0012",
 		"a/100",
-		"a~10",
 		"ab/1",
+		"a~10",
 		"b/1",
 		"b/01",
 		"b/2",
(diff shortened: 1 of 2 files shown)


niemeyer:
> if "a10" happens to be assigned to ar and "a1a" to br, that code would return true, while if the assignment was the other way around, it would return false.

That seems consistent and deterministic to me. What am I missing?

dominikgrygiel:
You're right. This PR has been premature then. I'll investigate further.

Issue #605, opened 2020-04-30
Support for user defined schemas and types mapping


pawel-kedzior-sw:
One of the coolest feature of yaml which I found when programming in other languages (C#/Java) was the ability to prepare a map of custom tags to concrete types and pass it to unmarshaler.
Example:

```
shapes:
   - !circle
      radius: 5
   - !square
      size: 2
```

By preparing a map of tags: '!circle' to a Circle type and '!square' to Square type, the 'shapes' field could be a list of AbstractShape (in Java/C#) or could be a slice of interface{} in Go.

The feature is equally applicable to single values, but the example with a collection is depicting it very well.

The feature is super useful because it makes the programming model very neat when you interpret the data.

As I tried to use the current yaml.v3, the best I could get is a map of properties. I couldn't also find out the details what was the original tag but I might have missed something.

I am requesting is to provide an API for unmarshaler to configure mapping in between the custom tags and concrete types and use it during unmarshaling to instantiate appropriate ones.

pawel-kedzior-sw:
I am working on the proposal of the pull-request here:

https://github.com/go-yaml/yaml/compare/v3...pawel-kedzior-sw:feature/custom-user-tags?expand=1

pawel-kedzior-sw:
Pull request: https://github.com/go-yaml/yaml/pull/636

NightTsarina:
This would be a very useful addition. In particular, I would like to support including files by using a custom schema (like https://github.com/tanbro/pyyaml-include does), and this seems to be the only missing piece.

Example:

```yaml
--- # main.yaml
myconfig: !include [conf.d/*.yaml]

<<: !include localoptions.yaml

--- # conf.d/foo.yaml
name: foo
option: bar

--- # conf.d/foo2.yaml
name: foo2
option: baz

--- # localoptions.yaml
extra: false
```

Which would unmarshal exactly as if the input was:

```yaml
myconfig:
- name: foo
  option: bar
- name: foo2
  option: baz
extra: false
```

PR #636, opened 2020-07-28
Feature/custom user tags

pawel-kedzior-sw:
This is pull request for improvement "Support for user defined schemas and types mapping" described in https://github.com/go-yaml/yaml/issues/605

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -1650,3 +1650,26 @@ func (s *S) TestFuzzCrashers(c *C) {
 //		yaml.Marshal(&v)
 //	}
 //}
+
+type Figure struct {
+	Shape interface{}
+}
+
+type Circle struct {
+	Radius int
+}
+
+func (s *S) TestCustomUserTags(c *C) {
+	var figure Figure
+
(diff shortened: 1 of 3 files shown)


decode_test.go
```diff
@@ -1650,3 +1650,72 @@ func (s *S) TestFuzzCrashers(c *C) {
...
+	Radius int
+}
+
+type Triangle struct {
+	Base int
+	Height int
+}
+
+type Figures struct {
+	Shapes []interface{}
```
thirsch, line 1668: In my current use case, it would also be nice to be able to define a custom interface for e.g. Circle and Triangle and being able to use this as the slice type in the Figures struct. Example:

```
type Circle struct {...}
func (Circle) Print() {...}

type Triangle struct {...}
func (Triangle) Print() {...}

type Shape interface {
  Print()
}

type Figures struct {
  Shapes []Shape
}
```

RangelReale:
This only supports custom tags for maps, maybe it should support  scalar types also?

PR #646, opened 2020-08-06
Fixed a bug that HeadComment could not output at the expected position when run Marshall.

WhiteRaven777:
Fixed an issue where the HeadComment was not outputting in the expected position.
A test for this fix has also been added.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module "gopkg.in/yaml.v3"
+module "gopkg.in/WhiteRaven777/yaml.v3"
 
 require (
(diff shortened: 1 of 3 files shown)


WhiteRaven777:
The following commitments are for me.
It has nothing to do with the content of this pull request, so please ignore it.
- 1f0a15b
  - fix go.mod

go.mod
```diff
@@ -1,4 +1,4 @@
-module "gopkg.in/yaml.v3"
+module "gopkg.in/WhiteRaven777/yaml.v3"
```
abursavich, line 1: oops

WhiteRaven777: sorry, thank you. 😂

PR #571, opened 2020-01-24
Feature (v2): set the default line length to infinity (-1)

laverya:
as suggested here: https://github.com/go-yaml/yaml/issues/166#issuecomment-555447882

PR #670, opened 2020-11-12
Revert "Set the default line length to infinity (-1) (#571)"

dims:
This reverts commit 0b1645d91e851e735d3e23330303ce81f70adbe3.

Hi,
At kubernetes, we are having a hard time with this change as it changes default behavior with no recourse to setting a custom width (to what it was before). Please see the following scenarios we hit recently:
- https://github.com/kubernetes/kubernetes/pull/95571#issuecomment-708796327
- https://github.com/kubernetes/kubernetes/pull/96425#issuecomment-725858135

Signed-off-by: Davanum Srinivas <davanum@gmail.com>

### DIFF
--- a/apic.go
+++ b/apic.go
@@ -86,7 +86,6 @@ func yaml_emitter_initialize(emitter *yaml_emitter_t) {
 		raw_buffer: make([]byte, 0, output_raw_buffer_size),
 		states:     make([]yaml_emitter_state_t, 0, initial_stack_size),
 		events:     make([]yaml_event_t, 0, initial_queue_size),
-		best_width: -1,
 	}
 }
 
--- a/encode_test.go
+++ b/encode_test.go
@@ -397,11 +397,6 @@ var marshalTests = []struct {
 		map[string]interface{}{"a": jsonNumberT("bogus")},
 		"a: bogus\n",
 	},
-	// Ensure that strings do not wrap
-	{
-		map[string]string{"a": "abcdefghijklmnopqrstuvwxyz ABCDEFGHIJKLMNOPQRSTUVWXYZ 1234567890 abcdefghijklmnopqrstuvwxyz ABCDEFGHIJKLMNOPQRSTUVWXYZ 1234567890 "},
-		"a: 'abcdefghijklmnopqrstuvwxyz ABCDEFGHIJKLMNOPQRSTUVWXYZ 1234567890 abcdefghijklmnopqrstuvwxyz ABCDEFGHIJKLMNOPQRSTUVWXYZ 1234567890 '\n",
-	},
 }
 
 func (s *S) TestMarshal(c *C) {


dims:
cc @laverya @niemeyer @jjshoe

jjshoe:
Why revert? Why not make it configurable, if that's what you need?

jjshoe:
If this is merged it will break all the code out there that's had work arounds removed.

jjshoe:
I would also suggest a test be added for this.

dims:
@jjshoe happy to the necessary work either way (add a test or change the patch to make it configurable). If you prefer the configurable path, what would that look like? (i.e, how can we specify the setting?)

laverya:
It'd look something like this, IMO: https://github.com/go-yaml/yaml/pull/455

(note that that PR was for go-yaml.v3, not v2 - the code is pretty much the same though)

bobbypage:
+1, https://github.com/go-yaml/yaml/pull/571 was a breaking change and is now causing issues for updating vendored dependencies in [github.com/kubernetes/kubernetes](http://github.com/kubernetes/kubernetes) due to tests expecting the old line length.

It would be great to either revert this breaking change or make it configurable.

jjshoe:
@bobbypage just as quick to update the tests over in kubernetes, no?

bobbypage:
I think there may be further reaching issues across kubernetes and it's consumers due to this change see https://github.com/kubernetes/kubernetes/pull/95571#issuecomment-709315815. @liggitt may have some more context there.

liggitt:
making it configurable seems useful, but preserving the default behavior of yaml.v2 prior to https://github.com/go-yaml/yaml/commit/0b1645d91e851e735d3e23330303ce81f70adbe3 is what I would expect

jjshoe:
Either way, if v2 is deprecated, and you move up to v3, you'll need to handle the change?

liggitt:
> Either way, if v2 is deprecated, and you move up to v3, you'll need to handle the change?

I would expect yaml.v3 to give configuration options to allow compatible output for yaml.v2 users, but yes, changing the default in yaml.v3 would be reasonable

niemeyer:
It was clearly a mistake to accept the formatting change in v2. The current situation is unfortunate, though, because if we remove the change we'll break some people twice to avoid breaking some people once. Given the size of Kubernetes both in terms of code and community, I'm willing to compromise on the side of reverting the change made to v2, but to reduce the pain of those that already made the migration towards the larger change, I'll also add an API that will enable the new behavior so that people can simply add a line to their code and not have to reformat everything again. I'll spend some time on this issue later today.

liggitt:
> to reduce the pain of those that already made the migration towards the larger change, I'll also add an API that will enable the new behavior so that people can simply add a line to their code and not have to reformat everything again

That seems like a good outcome. Thanks for looking at it.

dims:
@niemeyer sounds great! thanks for doing this. please let me know if i should close this PR.

niemeyer:
There it is: 7649d45

Please let me know if this looks reasonable and I'll add a version tag soon.

liggitt:
> There it is: [7649d45](https://github.com/go-yaml/yaml/commit/7649d4548cb53a614db133b2a8ac1f31859dda8c)
> 
> Please let me know if this looks reasonable and I'll add a version tag soon.

looks good to me, thanks

liggitt:
> looks good to me, thanks

Consumers still run the risk of poorly behaved things in their dependency tree toggling global options, but that's a general issue not specific to yaml or this option. The only thing I can think of to add would maybe be a comment indicating that global option should generally be set by the process owner (e.g. the `main()` implementer), not by a library

dims:
@niemeyer thanks a ton. +1 to "add a version tag"

dims:
i'll close this out as well.

niemeyer:
Thanks for the feedback. v2.4.0 is now tagged.

@liggitt That's a good point, but there's too much nuance to be able to easily provide guidance. Most proper software is split into independent packages internally, which may end up being the right place for that in some cases. Let's see what happens next. Hopefully v3 can soon take over and that issue will be history. I need to focus on the few open issues there and get it ready for a release.

Issue #681, opened 2020-12-18
v3: bug with parsing null values


rokkerruslan:
Version 3 differs from version 2 in its behavior when parsing null (or the equivalent `empty/~`) values. Version 2 had behavior conforms to the specification (partially), in particular clause `7.2. Empty Nodes': 

 > YAML allows the node content to be omitted in many cases. Nodes with empty content are interpreted as if they were plain scalars with an empty value. Such nodes are commonly resolved to a "null" value. 

The `v3` version has incorrect behavior because it considers `null` values as missing keys. Example:

```go
package main

import (
	"fmt"

	v2 "gopkg.in/yaml.v2"
	v3 "gopkg.in/yaml.v3"
)

type Config struct {
	Fields map[string]string `yaml:"fields"`
}

const configYAML = `
fields:
        f1: null
        f2: 
`

func main() {
	cfg := Config{}
	err := v2.Unmarshal([]byte(configYAML), &cfg)
	if err != nil {
		panic(err)
	}
	fmt.Println("v2", cfg.Fields)

	cfg = Config{}
	err = v3.Unmarshal([]byte(configYAML), &cfg)
	if err != nil {
		panic(err)
	}
	fmt.Println("v3", cfg.Fields)
}

```

Expected:

```
v2 map[f1: f2:]
v3 map[f1: f2:]

// Or parsing error
```

Got:

```
v2 map[f1: f2:]
v3 map[]
```

Go playground with example - https://play.golang.org/p/Sf_o_7ErXVN

--------------------------------------------------

The json parser from the standard library has a similar behavior to `v2` - [Playground](https://play.golang.org/p/B1ynxYuez0T)

niemeyer:
Thanks for the report. I'll have a look at this.

niemeyer:
I'm hoping the change in c8046fb addresses the intended use cases. Let me explain the situation and what was done.

In v2, we had a pretty straightforward rule of resetting whatever value was in place before to its zero value, which creates the behavior described for v2 in the original report above: the map was empty, and there are two null fields, so both of them get set to the zero value of the map which is the empty string in this case. Very easy to understand, but has an issue: default values.

In v3, that original behavior was changed to preserve the state of the value being decoded into when the data doesn't match the particular value type. Since the map value is a string, a null is rejected and is simply skipped, as any other incorrect value type would be.

[...]

PR #682, opened 2020-12-22
fixed behavior when parsing null values

i-pva:
I've tried to solve the issue (#681)  when parsing null values, removed type checking in switch block and made a default value for all scalar types.
I'll be glad for any opinion

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -548,11 +548,8 @@ func resetMap(out reflect.Value) {
 
 func (d *decoder) null(out reflect.Value) bool {
 	if out.CanAddr() {
-		switch out.Kind() {
-		case reflect.Interface, reflect.Ptr, reflect.Map, reflect.Slice:
-			out.Set(reflect.Zero(out.Type()))
-			return true
-		}
+		out.Set(reflect.Zero(out.Type()))
+		return true
 	}
 	return false
 }


i-pva:
maybe something went wrong with the tests.
all tests were passed on my local PC.

```
$ go test -v ./...

=== RUN   Test
OK: 45 passed
--- PASS: Test (4.85s)
=== RUN   ExampleUnmarshal_embedded
--- PASS: ExampleUnmarshal_embedded (0.00s)
PASS
ok      github.com/i-pva/yaml   (cached)
```

decode.go
```diff
@@ -548,11 +548,8 @@ func resetMap(out reflect.Value) {
 
 func (d *decoder) null(out reflect.Value) bool {
 	if out.CanAddr() {
-		switch out.Kind() {
-		case reflect.Interface, reflect.Ptr, reflect.Map, reflect.Slice:
-			out.Set(reflect.Zero(out.Type()))
-			return true
-		}
+		out.Set(reflect.Zero(out.Type()))
+		return true
```
niemeyer, line 552: Thank you very much for engaging and trying to fix that issue. Unfortunately, that's likely not what we want, as it changes the behavior and means nulls become valid options for scalars, unlike what we had and intended I believe. I'll have a look at it.

niemeyer:
Spoke too soon, sorry. It looks like that's almost exactly what we used to do in v2, and also what encoding/json does when it observes a null value. In addition, v2 also handled the special case of map that are not addressable, by zeroing them out.

I'm going to bring back that logic from v2.

niemeyer:
This issue is fixed as described in #681. The fix wasn't hard, but it was definitely subtle due to mentioned details.

i-pva:
Thanks you for your opinion. I got what you meant.  :+1:

PR #684, opened 2020-12-30
[v3] Improve comment handling

felixfontein:
Fixes two problems with comment handling in `v3` branch found while working on mozilla/sops#791:

1. When unmarshaling and re-marshaling
    ```.yaml
    a:
        b:
            # comment followed by newline

            c: d
    ```
    the comment jumps under `c: d`. (The new line is essential, without it it won't happen.)

2. When unmarshaling and re-marshaling
    ```.yaml
    # foo
    ```
    This parses to a zero node (the comment is lost).

3. When unmarshaling
    ```.yaml
    # foo
    ---
    key: value
    ```
    we end up with only one document containing both the comment and the hash.

The fix for 2 changes the parsing of "empty" documents which contain comments. Instead of returning `nil`, they now return a document node with no content.

I've added some unmarshal + re-marshal tests in `yaml_test.go`.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -202,17 +202,15 @@ func (p *parser) parseChild(parent *Node) *Node {
 }
 
 func (p *parser) document() *Node {
+	n := p.node(DocumentNode, "", "", "")
+	p.doc = n
 	if !p.event.implicit && len(p.event.head_comment) > 0 {
 		// This comment belongs to **before** the document event
-		n := p.node(DocumentNode, "", "", "")
-		// (1, 1) is not fully correct, but close enough
 		n.Line = 1
 		n.Column = 1
 		p.event.head_comment = nil
 		return n
 	}
-	n := p.node(DocumentNode, "", "", "")
-	p.doc = n
 	p.expect(yaml_DOCUMENT_START_EVENT)
 	p.parseChild(n)
 	if p.peek() == yaml_DOCUMENT_END_EVENT {
--- a/decode.go
+++ b/decode.go
@@ -141,7 +141,7 @@ func (p *parser) anchor(n *Node, anchor []byte) {
 	}
 }
 
-func (p *parser) parse() *Node {
+func (p *parser) parse(toplevel bool) *Node {
 	p.init()
 	switch p.peek() {
 	case yaml_SCALAR_EVENT:
@@ -156,6 +156,11 @@ func (p *parser) parse() *Node {
 		return p.document()
 	case yaml_STREAM_END_EVENT:
 		// Happens when attempting to decode an empty buffer.
+		if toplevel && len(p.event.head_comment) > 0 {
+			n := p.node(DocumentNode, "", "", "")
+			p.event.head_comment = nil
+			return n
+		}
 		return nil
 	case yaml_TAIL_COMMENT_EVENT:
 		panic("internal error: unexpected tail comment event (please report)")
@@ -191,7 +196,7 @@ func (p *parser) node(kind Kind, defaultTag, tag, value string) *Node {
 }
 
 func (p *parser) parseChild(parent *Node) *Node {
-	child := p.parse()
+	child := p.parse(false)
 	parent.Content = append(parent.Content, child)
 	return child
 }
--- a/decode.go
+++ b/decode.go
@@ -202,6 +202,15 @@ func (p *parser) parseChild(parent *Node) *Node {
 }
 
 func (p *parser) document() *Node {
+	if !p.event.implicit && len(p.event.head_comment) > 0 {
+		// This comment belongs to **before** the document event
+		n := p.node(DocumentNode, "", "", "")
+		// (1, 1) is not fully correct, but close enough
+		n.Line = 1
+		n.Column = 1
+		p.event.head_comment = nil
+		return n
+	}
 	n := p.node(DocumentNode, "", "", "")
 	p.doc = n
 	p.expect(yaml_DOCUMENT_START_EVENT)
--- a/emitterc.go
+++ b/emitterc.go
@@ -451,7 +451,6 @@ func yaml_emitter_emit_document_start(emitter *yaml_emitter_t, event *yaml_event
 			if !yaml_emitter_process_head_comment(emitter) {
 				return false
 			}
-			// Do not add newline if the document is empty
 			if !isEmpty && !put_break(emitter) {
 				return false
 			}
--- a/emitterc.go
+++ b/emitterc.go
@@ -359,6 +359,7 @@ func yaml_emitter_emit_stream_start(emitter *yaml_emitter_t, event *yaml_event_t
 func yaml_emitter_emit_document_start(emitter *yaml_emitter_t, event *yaml_event_t, first bool) bool {
 
 	if event.typ == yaml_DOCUMENT_START_EVENT {
+		isEmpty := emitter.events[emitter.events_head + 1].typ == yaml_DOCUMENT_END_EVENT
 
 		if event.version_directive != nil {
 			if !yaml_emitter_analyze_version_directive(emitter, event.version_directive) {
@@ -456,6 +457,9 @@ func yaml_emitter_emit_document_start(emitter *yaml_emitter_t, event *yaml_event
 		}
 
 		emitter.state = yaml_EMIT_DOCUMENT_CONTENT_STATE
+		if isEmpty {
+			emitter.state = yaml_EMIT_DOCUMENT_END_STATE
+		}
 		return true
 	}
 
--- a/emitterc.go
+++ b/emitterc.go
@@ -451,7 +451,8 @@ func yaml_emitter_emit_document_start(emitter *yaml_emitter_t, event *yaml_event
 			if !yaml_emitter_process_head_comment(emitter) {
 				return false
 			}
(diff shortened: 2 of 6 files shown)


felixfontein:
I had to revise my fix for 2, it now doesn't break anything else (that I know of). I've updated the description accordingly.

felixfontein:
I currently found some more problems and am looking at them.

felixfontein:
I think this is now ready.

felixfontein:
(I did sign the contributor agreement at https://ubuntu.com/legal/contributors and put @niemeyer as a reference.)

parserc.go
```diff
@@ -890,6 +893,14 @@ func yaml_parser_parse_block_mapping_value(parser *yaml_parser_t, event *yaml_ev
 	if token.typ == yaml_VALUE_TOKEN {
 		mark := token.end_mark
 		skip_token(parser)
+		// Move foot comment to head comment
```
niemeyer: I can see that's being done below, but I cannot see why this is being done. This comment is a nice place to describe that, as this is not an obvious operation.

decode.go
```diff
@@ -156,6 +156,11 @@ func (p *parser) parse() *Node {
 		return p.document()
 	case yaml_STREAM_END_EVENT:
 		// Happens when attempting to decode an empty buffer.
+		if toplevel && len(p.event.head_comment) > 0 {
```
niemeyer, line 159: Either the pre-existing comment above is incorrect, or we don't need the new variable. Also, it sounds like we should have enough state available in the parser to tell whether this is a top-level value or not. Can you provide some more details about what's going on here?

decode.go
```diff
@@ -191,12 +196,21 @@ func (p *parser) node(kind Kind, defaultTag, tag, value string) *Node {
...
+		// This comment belongs to **before** the document event
+		n := p.node(DocumentNode, "", "", "")
+		// (1, 1) is not fully correct, but close enough
+		n.Line = 1
+		n.Column = 1
+		p.event.head_comment = nil
+		return n
+	}
 	n := p.node(DocumentNode, "", "", "")
 	p.doc = n
```
niemeyer: Line 200 above is inside the if, and the following line should be there as well so that the behavior for the call-site remains the same. Positioning the if here would fix both.

decode.go
```diff
@@ -191,12 +196,21 @@ func (p *parser) node(kind Kind, defaultTag, tag, value string) *Node {
...
+	child := p.parse(false)
 	parent.Content = append(parent.Content, child)
 	return child
 }
 
 func (p *parser) document() *Node {
+	if !p.event.implicit && len(p.event.head_comment) > 0 {
+		// This comment belongs to **before** the document event
+		n := p.node(DocumentNode, "", "", "")
+		// (1, 1) is not fully correct, but close enough
```
niemeyer: Let's please just drop the comment here. Reading this code is clear that the lines are made up, and that's fine because the line is for the DocumentNode, not for the comment. Since the document is empty, there's nothing clearly wrong with this.

emitterc.go
```diff
@@ -450,12 +451,16 @@ func yaml_emitter_emit_document_start(emitter *yaml_emitter_t, event *yaml_event
 			if !yaml_emitter_process_head_comment(emitter) {
 				return false
 			}
-			if !put_break(emitter) {
+			// Do not add newline if the document is empty
```
niemeyer: This is just re-stating what we can read nicely in the line below.

yaml_test.go
```diff
@@ -0,0 +1,171 @@
+package yaml_test
```
niemeyer, line 1: We have a very nice set of tests for comments in the node_test.go file. These need to be integrated there as well, and the ones unrelated to comments can go into the respective test files.

yaml_test.go
```diff
@@ -0,0 +1,171 @@
...
+a:
+    ## foo
+    ##
+    b:
+`)
+}
+
+
+func (s *S) TestCommentEmptyDoc(c *C) {
+	testIdempotent(c, `# foo
```
niemeyer: That sort of test looks better as "#foo\n", etc.

niemeyer (commented):
Thanks for these proposed changes.  Here is an initial review. Please let me know how you feel about these points.

parserc.go
felixfontein: Improved in 193064d01c66d9643f3e6b9dd4142a2a83db62b4.

decode.go
felixfontein: Fixed in 193064d01c66d9643f3e6b9dd4142a2a83db62b4.

decode.go
felixfontein: Fixed in 193064d01c66d9643f3e6b9dd4142a2a83db62b4.

emitterc.go
felixfontein: Removed in 193064d01c66d9643f3e6b9dd4142a2a83db62b4.

yaml_test.go
felixfontein: Fixed in 193064d01c66d9643f3e6b9dd4142a2a83db62b4.

decode.go
felixfontein: I don't know / understand the parser well enough to be sure whether the comment is correct or not. That's why I added the parameter `toplevel`, to avoid accidentally introducing a bug here. If you prefer I can remove it.

yaml_test.go
felixfontein: I added most of the tests from this file (all single-document tests) to node_test.go as well.

felixfontein: (02f3190efc7e74bda69720bda4cbb88eea340de9)

decode.go
niemeyer: That's okay, but we cannot have additional logic in here unless we understand that well, otherwise we may be surprised in other situations, or may have misleading logic that makes it sound like this code executes at times it doesn't. Right now this is labeled as a "stream end event". Either that's true and the new logic is wrong, or it's incorrect and the old comment is wrong.

emitterc.go
```diff
@@ -359,6 +359,7 @@ func yaml_emitter_emit_stream_start(emitter *yaml_emitter_t, event *yaml_event_t
 func yaml_emitter_emit_document_start(emitter *yaml_emitter_t, event *yaml_event_t, first bool) bool {
 
 	if event.typ == yaml_DOCUMENT_START_EVENT {
+		isEmpty := emitter.events[emitter.events_head + 1].typ == yaml_DOCUMENT_END_EVENT
```
niemeyer, line 362: How do we know here that the +1 is safe?

yaml_test.go
niemeyer: Thanks, but we need to have all tests either in node_test.go or in the respective test files, using existing conventions. We have plenty of tests and test files already.

Also, let's please duplicate the exact same test in two different places. Those that are taken care by node_test.go don't need to exist again elsewhere.

parserc.go
```diff
@@ -890,6 +893,24 @@ func yaml_parser_parse_block_mapping_value(parser *yaml_parser_t, event *yaml_ev
...
+		// Move foot comment to head comment. This prevents that in some cases, comments
+		// in maps are moved to the wrong place. In the following map:
+		//
+		//     a:
+		//       b:
+		//         # comment followed by newline
+		//
+		//         c: d
+		//
+		// (the newline between the comment and the next line is essential!)
```
niemeyer, line 905: This is stated more succinctly in the body of the yaml right above, so no need to repeat here.

Otherwise, thanks for the comment. It provides clear rationale for the non-obvious logic.

niemeyer (changes requested):
Thanks for the tuning. I'm providing some additional comments on the PR below, but I also slept thinking of the underlying issue, and this feels like a partial solution to a more general problem.

Please hold on a moment... I'll provide a follow up comment after this review, and then will provide a candidate fix for the problem that maybe addresses the issue more generally.

niemeyer:
Okay, so this PR is doing two independent things, so the first thing I'd ask is let's please split those independent things apart, so that we can move both of them forward at their own pace.

## Issue one

So, first for the more tricky part of this PR: the current logic considers the first line after an arbitrary node to be a foot comment, as long as there's nothing further below it with no spacing that would justify it being its header. That logic fails when indentation visually associates the comment with the following data instead.

Your approach to fix this issue is a viable one: you're transforming what was a footer into a header. While this works, it special cases one particular instance in which we know this can happen, but there might be others that we need to look into. A more general approach is to fix the cause of the issue by not considering such cases as a foot comment in the first place.

That's what I've done in 3778e11. Please let me know how you feel about it, and whether it properly addresses cases you have at hand.

## Issue two

The second issue regards the handling of empty documents. That looks like you have a mostly clean implementation, except for the handling of the stream end event that is a bit uncertain per the comments in the reviews above.

Would you mind to split that apart, and also address the issues raised in terms of testing in the reviews?

Please let me know how you feel about these comments.

niemeyer:
I tuned it further to fix @mikefarah's #686 as well. This now works a bit closer to your PR since it looks at the token type, so I'm hoping that no use cases are left behind. I'm going to close this PR on that basis, but please me know if that's incorrect.

Also, we still need the empty document part of this PR. Let's please move this to a new PR so we can track that by itself.

PR #689, opened 2021-01-06
v3: Make UnmarshalYAML be called when values are set to null

dgl:
This potentially overlaps with #681, but it looks like the issue is elsewhere in the code, depending how that's fixed. I split the tests out into another commit so hopefully they'd be useful anyway.

A simple way to see this is at https://play.golang.org/p/LoO4LyKAwut -- change the import between v2 and v3 and watch how  the default field is not overwritten.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -397,17 +397,20 @@ func (d *decoder) callObsoleteUnmarshaler(n *Node, u obsoleteUnmarshaler) (good 
 // unmarshalling was already done by UnmarshalYAML, and if so whether
 // its types unmarshalled appropriately.
 //
-// If n holds a null value, prepare returns before doing anything.
+// If n holds a null value, prepare will not initialize any values, but will
+// call unmarshallers, in order to allow resetting default values to null.
 func (d *decoder) prepare(n *Node, out reflect.Value) (newout reflect.Value, unmarshaled, good bool) {
-	if n.ShortTag() == nullTag || n.Kind == 0 && n.IsZero() {
-		return out, false, false
-	}
+	orig := out
 	again := true
 	for again {
 		again = false
 		if out.Kind() == reflect.Ptr {
 			if out.IsNil() {
-				out.Set(reflect.New(out.Type().Elem()))
+				// Only if the value we have is not null, to avoid setting null fields
+				// to their default value.
+				if n.ShortTag() != nullTag && n.Kind != 0 && !n.IsZero() {
+					out.Set(reflect.New(out.Type().Elem()))
+				}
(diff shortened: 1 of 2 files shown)


decode.go
```diff
@@ -397,17 +397,20 @@ func (d *decoder) callObsoleteUnmarshaler(n *Node, u obsoleteUnmarshaler) (good
...
 		if out.Kind() == reflect.Ptr {
 			if out.IsNil() {
-				out.Set(reflect.New(out.Type().Elem()))
+				// Only if the value we have is not null, to avoid setting null fields
+				// to their default value.
+				if n.ShortTag() != nullTag && n.Kind != 0 && !n.IsZero() {
+					out.Set(reflect.New(out.Type().Elem()))
+				}
 			}
 			out = out.Elem()
```
niemeyer, line 416: This is only safe here because we're guaranteed to not be nil given the logic above. Once the above is conditional, this won't return a good out value. Am I missing something?

niemeyer (changes requested):
Thanks for the PR. The mentioned issue has been fixed, I hope, so this needs updating.

Also, although I understand the basic intention, I think we need to talk a bit about the proper intended behavior on edge cases. If we allow unmarshaling of null values, how do we get the value to initialize in the first place? It's not very helpful to call the function, and then throw away the result because the field was set to nil, which I think is what would happen with the suggested logic. So how does one say "this is actually nil so discard the value you created to call me" vs. "no, initialize and set this value as I wish", from within the method?

dgl:
The way I see it working is that if you have a struct like:

```go
struct {
  Value SomeType
}
```

Then that would always call SomeType's `UnmarshalYAML`, whereas if you have:

```go
struct {
  Value *SomeType
}
```

Then that would potentially set the pointer to nil (hence why I did the change in prepare() to attempt to follow pointers, although maybe that is wrong, will see about some more tests). I'll reply in more detail in #681 first though, as the behaviour outlined there is problematic for my use case.

decode.go
dgl: No and given your other refactor this can get a bit more simple too. I added some more tests too.

dgl:
I've added some tests (maybe a bit verbose) that show the behaviour I'd expect, i.e. when it's a pointer to something with an UnmarshalYAML it behaves differently to when it is not a pointer, as when there's no pointer involved it makes sense that UnmarhsalYAML gets to decide itself what the behaviour of null should be for the particular type.

harkishen:
Any plans on resuming the work on this PR?

Issue #460, opened 2019-05-08
KnownFields not supported in Node.Decode


tux21b:
The current API allows strict parsing into structs by setting ``Decoder.KnownFields(true)``. Elements can implement ``UnmarshalYAML(value *Node) error`` in order to influence the behavior of the decoding, and then can call ``node.Decode`` to get the default behavior. Unfortunately, ``node.Decode`` creates a new decoder internally without the known fields option enabled (the API silently looses the option) and there is no way to set it again.

anton-becker:
Moreover there is no way to check `KnownFields` value inside `UnmarshalYAML(value *Node) error`.

Emyrk:
Will this be addressed? It seems pretty serious as right now if you implement your own Unmarshal function, you lose this property. It is important for my application to keep it.

maja42:
It's been two years since the last update, but the issue still exists.
Is there any particular reason why this can't / hasn't been done yet?

drevell:
Here's a possible workaround. I'm not too familiar with Node/Kind/Content/etc so use with due caution.

```
type MyStruct struct {
	MyField `yaml:"my_field"`
}

func (s *MyStruct) UnmarshalYAML(n *yaml.Node) error {
	if err := extraFields(n, []string{"my_field"}); err != nil {
		return err
	}

	// do unmarshaling as normal ...
}

func extraFields(n *yaml.Node, knownFields []string) error {
	if n.Kind != yaml.MappingNode {
		return fmt.Errorf("got yaml node of kind %d, expected %d", n.Kind, yaml.MappingNode)
	}
	m := map[string]any{}
	if err := n.Decode(m); err != nil {
		return err
	}

	for k := range m {
		if !slices.Contains(knownFields, k) {
			return fmt.Errorf("unexpected field %s", k)
		}
	}
	return nil
}

```

Things I haven't considered that you might want to:
 - Efficiency for large numbers of fields, use set intersection or something
 - Efficiency again: everything gets unmarshaled twice
 - Position reporting: n.Line and n.Column will be the position of the *parent* yaml Node
 - Stringifying n.Kind rather than showing an integer in the error message
 - Use reflection to get the yaml name of each struct field so it doesn't have to be duplicated in the call to extraFields()

Issue #125, opened 2015-09-18
Unmarshaler interface doesn't work on inlined fields


wrouesnel:
So I've been trying to implement a config system where a lot of higher-order constructs inherit parameters from lower objects.

The problem I've found is that if I have two structs, Parent and Child, and Child contains Parent as an anonymous struct, then I can only get the fields of Child to unmarshal successfully if Parent does not have a custom UnmarshalYAML() function.

If a custom UnmarshalYAML() function is supplied, then only the nested structs fields will be populated - the top is set to nil.

Example - this works fine:

``` go
package main

import (
    "fmt"
    "gopkg.in/yaml.v2"
)

type Parent struct {
    DNA string          `yaml:"dna"`
}

type Child struct {
    Parent              `yaml:",inline"`
    Phenotype string    `yaml:"phenotype"`  
}

func main() {
    y := `
dna: AAAAGGGGTTTTCCCC
phenotype: Red Hair
`

    child := new(Child)
    err := yaml.Unmarshal([]byte(y),child)
    if err != nil {
        fmt.Println(err)
    }

    fmt.Println(child)
}
```

Result: `&{{AAAAGGGGTTTTCCCC} Red Hair}`

But this does not:

``` go
package main

import (
    "fmt"
    "gopkg.in/yaml.v2"
)

type Parent struct {
    DNA string          `yaml:"dna"`
}

func (this *Parent) UnmarshalYAML(unmarshal func(interface{}) error) error {
    type plain Parent
    if err := unmarshal((*plain)(this)); err != nil {
        fmt.Println("Parent :", err)    
        return nil
    }

    return nil
}

type Child struct {
    Parent              `yaml:",inline"`
    Phenotype string    `yaml:"phenotype"`  
}

func main() {
    y := `
dna: AAAAGGGGTTTTCCCC
phenotype: Red Hair
`

    child := new(Child)
    err := yaml.Unmarshal([]byte(y),child)
    if err != nil {
        fmt.Println(err)
    }

    fmt.Println(child)
}
```

Result: `&{{AAAAGGGGTTTTCCCC} }`

It also fails if skip the type indirection:

``` go
package main

import (
    "fmt"
    "gopkg.in/yaml.v2"
)

type Parent struct {
    DNA string          `yaml:"dna"`
}

func (this *Parent) UnmarshalYAML(unmarshal func(interface{}) error) error {
    tgt := make(map[string]interface{})

    if err := unmarshal(tgt); err != nil {
[...]

PR #691, opened 2021-01-15
DecodeWithOptions method for Node

osela:
Suggested fix for #460.

In my opinion it's the most straightforward way to fix the issue while keeping backwards compatibility and considering the small number of options (currently only one).

Obviously open for discussion.

### DIFF
--- a/node_test.go
+++ b/node_test.go
@@ -19,6 +19,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"reflect"
 
 	. "gopkg.in/check.v1"
 	"gopkg.in/yaml.v3"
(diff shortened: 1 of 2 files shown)


colega:
This would be great.

colega (approved):

colega:
@niemeyer is there any chance to get this merged? #460 is a blocker for us to adopt v3, while [v2 has bugs](https://github.com/go-yaml/yaml/issues/125) that require us to move away from it.

mitar:
I think this is nice and I would also love this to be merged in. The only issue left from https://github.com/go-yaml/yaml/issues/460 is that you cannot know inside `UnmarshalYAML` if `KnownFields` has been set or not, but that is a similar limitation to standard Go json package.

yxxhero:
@niemeyer WDYT? Thanks very much. it's a great PR.

fornellas:
Ping @niemeyer , is there anything missing for this to get merged? I'm up for even cutting a new PR as you prefer if that's the case, I'm really missing this feature.

PS: I doodled with an alternative interface here https://github.com/fornellas/yaml/tree/node_known_fields.

Issue #139, opened 2015-10-22
map[interface{}]interface{} on a map[string]interface{}


aanm:
Sorry to be persistent with this. I wasn't aware that yaml was able to have different types has keys than string. However, with the following example I see that `newstruct` is being considered has a `map[interface{}]interface{}`, is it a bug or is something I should expect?

```
type=string key=something type=string value=thing 
type=string key=newstruct type=map[interface {}]interface {} value=map[otherstruct:[map[foo-bar:12345]] bar:bar]
```

``` go
var data = `---
foo-map:
  something: "thing"
  newstruct:
    otherstruct:
      -
        foo-bar: 12345
    bar: "bar"`

type FooStruct struct {
    FooMap map[string]interface{} `yaml:"foo-map,omitempty"`
}

func main() {
    var foo FooStruct
    err := yaml.Unmarshal([]byte(data), &foo)
    if err != nil {
        fmt.Printf("error %+v", err)
    }
    for k, v := range foo.FooMap {
        fmt.Printf("type=%+v key=%+v type=%+v value=%+v\n", reflect.TypeOf(k), k, reflect.TypeOf(v), v)
    }
}
```

danprince:
I'm having trouble with this too. Even if I unmarshal into a `map[string]interface{}` the nested maps are of type `map[interface{}]interface{}`, meaning I can't then marshal to JSON (which must have string keys).

Is it possible to analyse the type of the interface passed to unmarshal then use the same type for recursively unmarshalling all the way down?

magnusbaeck:
> Is it possible to analyse the type of the interface passed to unmarshal then use the same type for recursively unmarshalling all the way down?

I'm not sure that would work in all cases. I also had this problem and made a custom unmarshaller in [github.com/elastic/beats:libbeat/common/mapstr.go](https://github.com/elastic/beats/blob/6435194af9f42cbf778ca0a1a92276caf41a0da8/libbeat/common/mapstr.go) that converts the keys of nested maps to strings so that the unmarshalled data can always be marshalled into JSON.

sontags:
The approach of @magnusbaeck works like a charm! I changed unmarshalYAML a bit as well as cleanUpMapValue in order to fit my needs (eg. it has now the same signature as yaml.Unmarshal() and ints, bool et al are not quoted since this works fine with encoding/json)... 

https://github.com/unprofession-al/gerty/blob/master/api/helpers.go

 It would be nice to see this in go-yaml as "compatibility mode".

[...]

PR #693, opened 2021-01-16
v3: Use map[string] for maps when merged keys are also strings

saracen:
The map type is `map[string]` when all keys in a map are strings. This adds an additional check to ensure that `map[string]` is also the map type used for merges, providing the map to be merged also only contains string keys.

This was reported by @nervo here: https://github.com/go-yaml/yaml/issues/139#issuecomment-566073291

For example:
```
foo:
    &foo
    bar: baz

bar:
    <<: *foo
    baz: qux
```

At the moment returns a `map[interface{}]interface{}`, when ideally, it should be a `map[string]interface{}` because both `bar` and `foo` contain only string keys. This PR fixes this.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -844,7 +844,18 @@ func isStringMap(n *Node) bool {
 	}
 	l := len(n.Content)
 	for i := 0; i < l; i += 2 {
-		if n.Content[i].ShortTag() != strTag {
+		switch n.Content[i].ShortTag() {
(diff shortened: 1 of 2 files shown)


saracen:
@niemeyer are you happy with this approach?

`TestMerge` is simplified, because all of the keys are strings. I've added additional tests for when they're not.

PR #694, opened 2021-01-18
Update gopkg.in/check.v1 to latest version

cfergeau:
There has been a few commits to this module in the last year, using a
version from 2020 rather than from 2016 should not hurt, and makes it
look less like a stale dependency.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,10 @@
-module "gopkg.in/yaml.v3"
+module gopkg.in/yaml.v3
+
+go 1.16
+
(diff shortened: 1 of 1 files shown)


Abirdcfly:
any update here ? 🧐

cfergeau:
Just updated the PR to pick up a version change to the indirect `github.com/kr/pretty` dependency, and added a `go 1.16` version at the beginning as latest go adds one by default. But it would indeed be nice to get this PR in :)

PR #690, opened 2021-01-07
[v3] Improve empty document handling

felixfontein:
Parts of #684 that haven been fixed by @niemeyer in the meantime.

> Thanks, but we need to have all tests either in node_test.go or in the respective test files, using existing conventions. We have plenty of tests and test files already.

So far I found exactly one test for multi-document reading and writing, namely `TestEncoderMultipleDocuments` in `encode_test.go`. Where should I put the multi-document decoding and encoding tests with `Node`?

PR #697, opened 2021-01-30
[v3] Fix CI

felixfontein:
CI in https://github.com/go-yaml/yaml/pull/690 currently fails with `Error: Unable to find Go version '1.16.0-beta1' for platform linux and architecture x64.`.


felixfontein:
Meh. Would have been too easy :)

PR #706, opened 2021-02-26
Update go.yaml

ryuichi1208:
Fixed the version of go used in CI


felixfontein:
@niemeyer would be great if this can be merged, that would fix CI!

Issue #709, opened 2021-03-09
v3: preserve comments lines and indents after unmarshal -> marshal


Unix4ever:
I wonder if it is possible to preserve the same positions for all decoded comments, when you unmarshal into `yaml.Node` and then marshal that `yaml.Node` back to yaml.

So for example:

```yaml
 block1:
    # the comment
    map:
        key1: a
        key2: b
    
    # foot comment
block2:
    hi: there
```

Turns into:
```yaml
 block1:
    # the comment
    map:
        key1: a
        key2: b
        # foot comment
block2:
    hi: there
```

Where foot comment changes it's indent and becomes foot comment of key2 (?).

Here's the code snippet:

```go
package main

import (
	"log"
	"strings"
	
	yaml "gopkg.in/yaml.v3"
)

func main() {
	var node yaml.Node
	
	data := []byte(strings.TrimSpace(`
block1:
    # the comment
    map:
        key1: a
        key2: b
    
    # foot comment
block2:
    hi: there

`))

	log.Printf("INPUT:\n %s", data)

	if err := yaml.Unmarshal(data, &node); err != nil {
		log.Fatalf("Unmarshalling failed %s", err)
	}
	
	results, err := yaml.Marshal(node.Content[0])
	if err != nil {
		log.Fatalf("Marshalling failed %s", err)
	}
	
	log.Printf("RESULT:\n %s", results)
}
```

https://play.golang.org/p/BtzerXV8ofI
It looks like the comment is attached to the last element in the map, instead of attaching it to the map.

Do you think it's a bug in the library? Or that's intended behavior? Is there any workaround for that?

niemeyer:
On a first sight this seems to be a bug. It should remain as a foot comment of the key, otherwise it'll keep walking down the hierarchy on future encode/decode cycles. I'll have a look at this on the next update.

Thanks for the report.

wburningham:
@niemeyer how big of a lift is it to fix this bug? I also have seen this issue.

rafaela-soares:
Is there any update about this issue? I am facing a similar problem.

Original file content:
```
---
# this playbook creates users and projects in openstack,
# it's used for a "local authentication" mode, not for SSO

- name: Playbook to Create users and Projects in Openstack
  hosts: localhost
  collections:
    - openstack.cloud
  tasks:
    - name: 'Ensure Projects are as defined'
      include: subroutines/openstack_per_project_actions.yaml

    - name: 'Create Users in Openstack'
[...]

PR #710, opened 2021-03-12
Fix indents of the foot comments going after the map

Unix4ever:
Removed the function that was adding the comments to the last key from
the map and replaced that with the separate comments queue.

Unfold that comment queue each time we get the end of a flow, a document and
a block.
Append comments to the end token if their indents are >= than the first
comment indent (this is safe as we only add comments to that queue if we
detect that their indent is < than the last key indent).

Additionally fixed the bug in the emitter, related to skipping
foot_comments for the mapping end events.

Fixes: https://github.com/go-yaml/yaml/issues/709

Signed-off-by: Artem Chernyshev <artem.0xD2@gmail.com>

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -296,10 +296,6 @@ func (p *parser) mapping() *Node {
 	}
 	n.LineComment = string(p.event.line_comment)
 	n.FootComment = string(p.event.foot_comment)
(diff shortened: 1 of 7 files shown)


Unix4ever:
ugh, no I kinda failed here, this PR still has many issues with comments.

PR #711, opened 2021-03-15
Introduce RawYAML interface that can allow passing raw yaml blocks

Unix4ever:
Then these block will be inlined as valid YAML istead of using quoted or
escaped format which is used for strings.
This can be used if you already have a decoded yaml which you want to inline into another object, when decoding/encoding that yaml is not really option.
That is helpful to get a workaround for comments decoding issues like https://github.com/go-yaml/yaml/issues/709

Signed-off-by: Artem Chernyshev <artem.0xD2@gmail.com>

### DIFF
--- a/apic.go
+++ b/apic.go
@@ -1,17 +1,17 @@
-// 
+//
 // Copyright (c) 2011-2019 Canonical Ltd
 // Copyright (c) 2006-2010 Kirill Simonov
-// 
+//
 // Permission is hereby granted, free of charge, to any person obtaining a copy of
 // this software and associated documentation files (the "Software"), to deal in
 // the Software without restriction, including without limitation the rights to
 // use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
 // of the Software, and to permit persons to whom the Software is furnished to do
 // so, subject to the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be included in all
 // copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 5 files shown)


lastthyme (approved):

niemeyer:
Thanks for the PR, but unfortunately we cannot do as-is bytes with YAML as it's context-sensitive. Please see #13 for a conversation on the topic.

smira:
@niemeyer we were looking for this change to preserve YAML document as is, with comments and complex structure while embedding that into another YAML document as a mapping value. it works fine in terms of indentation etc. 

it might not be the best approach, but unfortunately Unmarshal/Marshal cycle messes up the comments bad enough in the doc rendering it unreadable.

Unix4ever:
Initially I tried to address the issue with comments decoding/encoding not being idempotent:
See: https://github.com/go-yaml/yaml/issues/709

I got some progress there. But it started to break other parts of the comments encoder.
And then I also found more idempotency issues related to comments encoding, so I don't think it was worth it.

Fixing all comment encoding issues may work for us, but it's way more work than having raw YAML blocks.

niemeyer:
@smira I'm keen on learning about such cases where the document is rendered unreadable. The "raw" idea is not a reasonable approach because it cannot be raw. Unlike JSON, YAML depends both on indentation but also on other details of the surrounding environment, such as flow style. So the "raw" block will need to be parsed and interpreted, and then a new can of worms is open.

The best alternative for now is indeed yaml.Node in v3, and if that's broken somehow we need to fix it either way.

niemeyer:
@Unix4ever Raw YAML blocks are not less work because they're inherently not a feasible idea for YAML, as described above. If the idea is just to have a dumb text block that could be breaking the YAML logic depending on what's inside it and what's outside it, then we don't need a YAML encoder/decoder for that: text/template would do.

Thanks for the issue related to the comments formatting. I'll respond there so we can have a single thread on it.

PR #733, opened 2021-04-27
Fix key-sorting bugs.

tomclegg:
Existing key sorting is nondeterministic because (keyList)Less() returns contradictory results.
* Less("7a","8a") returns true: 7 < 8
* Less("8a","77") returns true: 8 < 77
* Less("77","7a") returns true: the first char is equal ("continue"), then "7" < "a" because "a" is a letter

(This has been [reported previously on a closed issue](https://github.com/go-yaml/yaml/pull/195#issuecomment-440818560).)

There are other surprising behaviors, too:
* Numbers that overflow int64 are sorted strangely ("9223372036854775808" < "9223372036854775807")
* Non-ASCII digits are sorted strangely ("\u0666" is sorted as if it's "1590")

This commit fixes the above issues, adds some keys in the sorting test to expose some edge cases, and adds a loop in the test to more reliably catch nondeterministic sorting.

### DIFF
--- a/encode_test.go
+++ b/encode_test.go
@@ -614,30 +614,56 @@ func (s *S) TestSortedOutput(c *C) {
 		"d7abc",
 		"d12",
 		"d12a",
+		"d77",
+		"f,0",
+		"f,00",
+		"f,000a",
+		"f,01",
+		"f,001",
+		"f,001a",
+		"f,0001a",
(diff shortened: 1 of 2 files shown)


tomclegg:
Existing unmerged PR #439 addresses the nondeterministic behavior but not the int64/non-ascii issues.

rabbbit:
@niemeyer (I think you're the owner of this repo?) - would you be able to take a look? We hit a sorting problem recently.

PR #746, opened 2021-06-09
v3: Don't count sequence indicator as part of indentation.

monopole:
This PR removes 2 spaces from sequence indentations to account for space consumed by the sequence item indicator, and adds tests around the use of the `SetIndent` api.

This maintains sequence indentation consistency with gopkg.in/yaml.v2 (see the example at https://pkg.go.dev/gopkg.in/yaml.v2#readme-api-documentation) and conforms to the comment:

> both the “-” indicator and the following spaces are considered
> to be part of the indentation of the nested collection.

in the specification at https://yaml.org/spec/1.2/spec.html#id2772075

Kubernetes is currently pinned to  gopkg.in/yaml.v2, and kustomize (now a part of kubectl) is pinned to v3.  We'd like the marshalling of sequences by kustomize and kubectl (e.g. `kubectl create deployment foo -o yaml --dry-run --image bar`) to match.  This means reverting the parts of https://github.com/go-yaml/yaml/commit/ae27a744346343ea814bd6f3bdd41d8669b172d0 and  https://github.com/go-yaml/yaml/commit/a5ece683394c3b88d90572e44d36c93aea492c2c that dealt with sequence indentation.  

There's more background in https://github.com/kubernetes-sigs/kustomize/issues/3946.  There's time pressure as we'd like kubectl and kustomize to be consistent before the next code kubernetes code freeze in early July, and this PR or a fork seem like the only options.

### DIFF
--- a/emitterc.go
+++ b/emitterc.go
@@ -225,6 +225,15 @@ func yaml_emitter_append_tag_directive(emitter *yaml_emitter_t, value *yaml_tag_
 	return true
 }
 
+// Don't count the sequenceIndicator as part of sequence indentation.
+// This maintains sequence indentation consistency with gopkg.in/yaml.v2
+// (https://pkg.go.dev/gopkg.in/yaml.v2#readme-api-documentation)
+// and conforms to the YAML 1.2 spec comment
+//   > both the “-” indicator and the following spaces are considered
+//   > to be part of the indentation of the nested collection.
+// at https://yaml.org/spec/1.2/spec.html#id2772075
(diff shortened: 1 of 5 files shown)


monopole:
This indentation reversion should also help with moving kubernetes-sigs/yaml to v3, see https://github.com/kubernetes-sigs/yaml/issues/18

monopole:
Abandoning this for now; the change always resulted in a few bad failures of existing or new tests.

PR #750, opened 2021-06-17
added a flag to indent or not indent the sequence

### LINKED ISSUE #720: v3 - Configurable or autodetected list indentation
Hi!

When using v3, identation on list child are not preserved. That was the case in v2.

Input:
```
list:
- foo: bar
```

Expected:
```
list:
- foo: bar
```

Actual (based on encoder indent):
```
list:
    - foo: bar
```

mikebz:
fixes #720

This provides a function where one can set whether to indent or not indent the sequence.  Not indenting the sequence provides a smoother upgrade path from v2 to v3.  A lot of tools (like kustomize) would break their clients if the expected yaml differed.

If we expect more of these types of output configurations we should consider creating a formatting flags structure.  Right now we just have the indent size and this flag, but maybe it will grow in the future.  Happy to do either one, but my guess is that right now this is a stop gap for updates.

### Review needed
1. @niemeyer - does this align with your approach?
1. Should we add more tests?  Any cases that are crucial to cover?


.gitignore
```diff
@@ -0,0 +1 @@
+.vscode
```
natasha41575, line 1: why add this?

encode_test.go
```diff
@@ -656,6 +656,18 @@ func (s *S) TestSetIndent(c *C) {
 	c.Assert(buf.String(), Equals, "a:\n        b:\n                c: d\n")
 }
 
+// This test is providing coverage for SetIndentSequences function.
+// When we are setting that to false the sequences are not indented.
+func (s *S) TestSetIndentSequences(c *C) {
+	var buf bytes.Buffer
+	enc := yaml.NewEncoder(&buf)
+	enc.SetIndentSequences(false)
```
natasha41575, line 664: should there be a test for setting it to true?

.gitignore
mikebz: VS Code is a popular editor.  I was surprised that there was not a gitignore for this.

encode_test.go
mikebz: yes, it's actually true (well the inverse under the hood) by default so the above tests for sequences cover it.  We can add an explicit set to true and make sure it works.  It's not a bad idea.

mikebz:
Superceded by @natasha41575 's https://github.com/go-yaml/yaml/pull/753

PR #753, opened 2021-06-22
compact seq indent option

### LINKED ISSUE #720: v3 - Configurable or autodetected list indentation

natasha41575:
fixes #720, needed for kustomize to avoid breaking changes when upgrading our yaml v3 dependency. 

This PR also includes a needed fix for https://github.com/go-yaml/yaml/issues/755 in its last commit, but I can remove it if preferred. The commit in https://github.com/go-yaml/yaml/pull/756 is the same as the last commit in this PR. 

Happy to update the PR to expose the flag in a different way if that's desired. 

Same questions as #750:
1. @niemeyer - does this align with your approach?
2. Should we add more tests? Any cases that are crucial to cover?


go.sum
```diff
@@ -0,0 +1,4 @@
+gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405 h1:yhCVgyC4o1eVCa2tZl7eS0r+SDo693bJlVdllGtEeKM=
```
mikebz: I noticed that @niemeyer didn't have a go.sum.  It might be that the build or whatever else produces it.  I would not introduce this file, it seems like an unrelated change.

yaml.go
```diff
@@ -278,6 +278,11 @@ func (e *Encoder) SetIndent(spaces int) {
 	e.encoder.indent = spaces
 }
 
+// DisableSeqIndent disables the indentation for sequence items.
+func (e *Encoder) DisableSeqIndent() {
+	e.encoder.emitter.disable_sequence_indent = true
```
mikebz: is there a way this can be called prior to the encoder and emitter being instantiated?

go.sum
natasha41575: done

yaml.go
natasha41575: good question, though I'm wondering why we would want to do that?

mikebz: I don't think we wold want to do that, but in general as a package designer one should anticipate that people will do things that don't make sense.  Is there a way to run into trouble with this interface?  Can we avoid it?  My PR had a flag on the current struct and then set one on the internal struct at init().

I guess it's not a huge deal and we can see if people break themselves first.

natasha41575:
Still interested in seeing this merged, but closing due to lack of activity. Please let me know if you have any interest in continuing this discussion.

Issue #755, opened 2021-06-25
v3: leading newline gets removed


natasha41575:
When writing a file such as 
```

a:
  b:
    c: d
```

The newline at the top of the file is not preserved. It appears that this only happens with the first leading newline; if there are multiple leading newlines only one is removed. 

This started happening after https://github.com/go-yaml/yaml/commit/496545a6307b2a7d7a710fd516e5e16e8ab62dbc; prior to this commit the leading newline was preserved correctly.

natasha41575:
https://github.com/go-yaml/yaml/pull/753 has a fix for this in its last commit.

I have also made a separate PR https://github.com/go-yaml/yaml/pull/756 to add a test and address this issue specifically.

PR #756, opened 2021-06-25
fix leading newline issue

natasha41575:
fixes https://github.com/go-yaml/yaml/issues/755

The last commit in https://github.com/go-yaml/yaml/pull/753 is the same as this PR.


inteon:
@niemeyer can this bugfix get merged?

PR #760, opened 2021-06-30
Allow to get source position of unmarshal error

### LINKED ISSUE #759: Feature request: Unmarshal error with source position
Hi,

I'm using this library to develop [actionlint](https://github.com/rhysd/actionlint). Since it is a linter, I want to know the exact source position of unmarshal error so that it can report exact error position with helpful source snippet.

Currently go-yaml/yaml's unmarshal error does not have source position information though line number is embedded in an error message.

Example usage is as follows:

```go
package main

import (
	"fmt"

	"gopkg.in/yaml.v3"
)

func main() {
	in := "#\n - - QI\xd7"

	var n yaml.Node
	if err := yaml.Unmarshal([]byte(in), &n); err != nil {
		err, _ := err.(*yaml.TypeError)
		for _, err := range err.Errors {
			// I want the err.Line and err.Column fields
			fmt.Fprintf("yaml parse failure at line:%d col:%d: %s\n", err.Line, err.Column, err.Message)
		}
		fmt.Fprintf("%d errors found while parsing yaml file\n", len(err.Errors))
	}
}
```

rhysd:
Fixes #759

With this PR, error instances in `TypeError.Errors` know their source positions. Previously the field was string slice, but now it is slice of struct of message, line and column so that the caller can know where each error occurred exactly.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -310,7 +310,7 @@ func (p *parser) mapping() *Node {
 type decoder struct {
 	doc     *Node
 	aliases map[*Node]bool
-	terrors []string
+	terrors []UnmarshalError
 
 	stringMapType  reflect.Type
 	generalMapType reflect.Type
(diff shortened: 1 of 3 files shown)


andig:
It would be absolutely fantastic to have this capability. 

/cc @niemeyer is this blocked in any way? Could a small sponsoring help to move this forward?

andig:
@rhysd I think there is a second case that's not handled. When `decode` panics, it moves into `handleErr`:

```
func handleErr(err *error) {
	if v := recover(); v != nil {
		if e, ok := v.(yamlError); ok {
			*err = e.err
		} else {
			panic(v)
		}
	}
}
```

Does `yamlError` need line information, too?

rhysd:
It would be necessary. But I don't have enough time to update this branch for now.

andig:
@rhysd Maybe you could take a look- forked your PR and added terminal failures with line numbers.

rhysd:
Looks good to me. Thank you for improving this branch. I'm closing this in favor of #901.

PR #774, opened 2021-08-06
yaml: make struct info tag configurable

martinrode:
In order to parse Kubernetes resources, we need to use the json struct info tags provided by Kubernetes.

### DIFF
--- a/yaml.go
+++ b/yaml.go
@@ -545,7 +545,7 @@ func getStructInfo(st reflect.Type) (*structInfo, error) {
 
 		info := fieldInfo{Num: i}
 
-		tag := field.Tag.Get("yaml")
+		tag := field.Tag.Get(YAML_STRUCT_TAG)
 		if tag == "" && strings.Index(string(field.Tag), ":") < 0 {
 			tag = string(field.Tag)
 		}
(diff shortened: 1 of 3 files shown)


yaml.go
```diff
@@ -545,7 +545,7 @@ func getStructInfo(st reflect.Type) (*structInfo, error) {
 
 		info := fieldInfo{Num: i}
 
-		tag := field.Tag.Get("yaml")
+		tag := field.Tag.Get(YAML_STRUCT_TAG)
```
mengqiy, line 548: I encounter the same problem when parsing k8s resources.
Does it make sense to make it fall back on json tags if yaml tags are not available?

unmarshall:
**Observation:**
Most of the k8s API structs have `json` tags. Wanting to change that to add `yaml` tags is an uphill task. This library will have limited utility if it cannot marshall/unmarshall json tags as k8s has already become the de-facto container orchestration engine.
I agree to the arguments that JSON is not YAML. So there are only 2 options - stick to this stand and then force people to move to other libraries or be a bit flexible as YAML is anyways a superset of JSON and allow fallback to JSON tags. 

Unless a general change is made to not have json/yaml/toml etc specific tags to specify names for variables when they are serialized or de-serialized we will have to live with this oddity to specify multiple tags for different file formats duplicating the same name everywhere. This is a far fetched dream which might come true in future (sure but no one can say when).

PR #792, opened 2021-10-11
feat: add inline struct ptr support

### LINKED ISSUE #55: Option ,inline needs a struct value field
Is there any reason against inlining member being pointer to a struct, complementary to a struct value?

No issue with extending the switch [here](https://github.com/go-yaml/yaml/blob/v2/yaml.go#L258)?

stevenh:
Add support for inline struct pointers for exported fields.

Also:
* Cleanup errors using fmt.Errorf instead of string concatenating.
* Removed commented out code.
* Formatted edited files with goimports.

Fixes #55

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -762,7 +762,7 @@ func (d *decoder) mappingStruct(n *node, out reflect.Value) (good bool) {
 			if info.Inline == nil {
 				field = out.Field(info.Num)
 			} else {
-				field = out.FieldByIndex(info.Inline)
+				field = fieldByIndex(out, info.Inline)
 			}
 			d.unmarshal(n.children[i+1], field)
 		} else if sinfo.InlineMap != -1 {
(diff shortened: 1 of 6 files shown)


corpix:
Lack of feature like this breaks Go's semantics. Because you can set embedded struct fields like they are defined on the struct itself.

```go
type Foo struct {
  *Bar
}
type Bar struct {
  Blah string
}

func (f *Foo) Default() {
  f.Blah = "hello"
}
...
```

stevenh:
Looks like you have an [additional fix](https://github.com/corpix/yaml/commit/faaebb51ad1f6e1c53f6eca0f406bd9f4f898213) there @corpix, should I incorporate that into this PR?

corpix:
Oops, missed this mention!
@stevenh shore, feel free to reuse this code as you like.

stevenh:
This is now ready for review, thanks to @corpix for the panic fix.

hellt:
Hi @niemeyer 
Has it been given any consideration given that the PR is available?

drgrib:
Still interest in this PR :) I get the low priority though, since I can just make my embedded type a non-pointer and it works.

stevenh:
Still a useful feature IMO

PR #794, opened 2021-10-12
fix: inline map set to nil

### LINKED ISSUE #793: inline maps get set to nil
If a struct contains an inline map it gets set to nil if no values are present.
Example:
```go
package main

import (
	"fmt"
	"log"

	"gopkg.in/yaml.v2"
)

type MyStruct struct {
	Name   string
	Inline `yaml:",inline"`
}

type Inline map[string]interface{}

func main() {
	v := &MyStruct{
		Name:   "test",
		Inline: Inline{"key": "val"},
	}
	fmt.Printf("%#v\n", v)
	err := yaml.Unmarshal([]byte(`{name: "wibble"}`), v)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Printf("%#v\n", v)
}
```
Output:
```go
&main.MyStruct{Name:"test", Inline:main.Inline{"key":"val"}}
&main.MyStruct{Name:"wibble", Inline:main.Inline(nil)}
```

Expected:
```go
&main.MyStruct{Name:"test", Inline:main.Inline{"key":"val"}}
&main.MyStruct{Name:"wibble", Inline:main.Inline{"key":"val"}}
```

[Playground link.](https://play.golang.org/p/KJywGxmh1iK)

The bug is the call to [inlineMap.Set(reflect.New(inlineMap.Type()).Elem())](https://github.com/go-yaml/yaml/blob/v2/decode.go#L736)

stevenh:
Fix line maps being set to nil if no elements are found after calling yaml.Unmarshal.

Fixes #793

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -733,7 +733,6 @@ func (d *decoder) mappingStruct(n *node, out reflect.Value) (good bool) {
 	var elemType reflect.Type
 	if sinfo.InlineMap != -1 {
 		inlineMap = out.Field(sinfo.InlineMap)
-		inlineMap.Set(reflect.New(inlineMap.Type()).Elem())
 		elemType = inlineMap.Type().Elem()
 	}
 


stevenh:
Cycling back to aging PR's so wanted to check if this package still being maintained @niemeyer ?

PR #752, opened 2021-06-18
Add a function to disable unique keys

xyz-li:
Provide function `DisableUniqueKeys` to disable unique keys option.

Fixes #751

PR #796, opened 2021-10-26
add support for strict errors

inteon:
based on #752

Now, "unknown field" errors are considered strict errors.
This means that unmarshalling will still return a valid response, but a strict error is returned too.
The library user can now decide eg. to use the result and also display a warning.

ref: https://github.com/kubernetes-sigs/yaml/pull/61


decode.go
```diff
@@ -308,9 +308,10 @@ func (p *parser) mapping() *Node {
 // Decoder, unmarshals a node into a provided value.
 
 type decoder struct {
-	doc     *Node
-	aliases map[*Node]bool
-	terrors []string
+	doc           *Node
+	aliases       map[*Node]bool
+	terrors       []string
+	strictTerrors []string
```
luxas: this should probably be `strictErrors`? I guess t stands for type, but it is still a bit weird when reading

decode.go
```diff
@@ -870,10 +869,6 @@ func (d *decoder) mappingStruct(n *Node, out reflect.Value) (good bool) {
 		d.prepare(n, field)
 	}
 
-	var doneFields []bool
-	if d.uniqueKeys {
-		doneFields = make([]bool, len(sinfo.FieldsList))
```
luxas: Why is this (and below) completely removed; I would have expected `terrors` to just swapped to the strict variant?

decode.go
```diff
@@ -886,13 +881,6 @@ func (d *decoder) mappingStruct(n *Node, out reflect.Value) (good bool) {
 			continue
 		}
 		if info, ok := sinfo.FieldsMap[name.String()]; ok {
-			if d.uniqueKeys {
-				if doneFields[info.Id] {
-					d.terrors = append(d.terrors, fmt.Sprintf("line %d: field %s already set in type %s", ni.Line, name.String(), out.Type()))
```
luxas: I would have expected terrors to just swapped to the strict variant?

luxas (commented):
I'm still not convinced (more context in thread https://github.com/kubernetes-sigs/yaml/pull/61) that we should allow yaml.v3 to be non-spec compliant and allow duplicate fields at any time, although yaml.v2 allows duplicate fields by default.

JSON just "recommends" unique keys, for which it then makes sense that unique keys is a sense of strictness, but YAML mandates their uniqueness, hence I do not think of that as a strictness measure.

I think the change to separate strict and non-strict errors is good though.

decode.go
inteon: changed it

inteon:
@luxas as requested, I now only consider "unknown field" errors as strict errors.

yaml.go
```diff
@@ -129,8 +129,8 @@ func (dec *Decoder) Decode(v interface{}) (err error) {
 		out = out.Elem()
 	}
 	d.unmarshal(node, out)
-	if len(d.terrors) > 0 {
-		return &TypeError{d.terrors}
+	if len(d.terrors) > 0 || len(d.strictErrors) > 0 {
```
luxas, line 132: not sure if this is now common enough to have its own helper func attached to the decoder, something like `if err := d.checkErr(); err != nil { return err }` ... it would add some overhead due to one extra fn call so not sure if it's worth it.

luxas (approved):
Thanks! The patch LGTM.

@liggitt @niemeyer do you have further comments or is this ok? Thanks for your time :100:!

inteon:
I'm closing this PR, since I'm currently working on other things.

PR #798, opened 2021-11-07
WIP: YAML Testsuite Tester interface

pantoniou:
Implement a rudimentary Tester interface that allows generation
of YAML testsuite textual representation of an internal event.

Please see: https://github.com/yaml/yaml-test-suite

### DIFF
--- a/yaml.go
+++ b/yaml.go
@@ -815,7 +815,7 @@ func (tst *Tester) NextEvent() *string {
 		}
 		str += escaped(e.value)
 	case yaml_ALIAS_EVENT:
-		str = "=ALI *" + string(e.value)
+		str = "=ALI *" + string(e.anchor)
 	default:
 		panic("internal error: Unexpected event: (please report): " + p.event.typ.String())
 	}
--- a/yaml.go
+++ b/yaml.go
@@ -761,7 +761,7 @@ func (tst *Tester) NextEvent() *string {
 		str = "-STR"
 		tst.got_strend = true
 	case yaml_DOCUMENT_START_EVENT:
-		if !e.implicit {
+		if e.implicit {
 			str = "+DOC"
 		} else {
 			str = "+DOC ---"
--- a/yaml.go
+++ b/yaml.go
@@ -696,3 +696,132 @@ func isZero(v reflect.Value) bool {
 	}
 	return false
 }
+
+func TestPanto(v int) {
+	fmt.Printf("TestPanto v=%d\n", v)
+}
+
+type Tester struct {
+	parser	*parser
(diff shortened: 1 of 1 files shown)


yaml.go
```diff
@@ -696,3 +696,132 @@ func isZero(v reflect.Value) bool {
...
+	case yaml_STREAM_START_EVENT:
+		str = "+STR"
+	case yaml_STREAM_END_EVENT:
+		str = "-STR"
+		tst.got_strend = true
+	case yaml_DOCUMENT_START_EVENT:
+		if !e.implicit {
+			str = "+DOC"
+		} else {
+			str = "+DOC---"
```
perlpunk: Shouldn't there be a space before the `---`?

yaml.go
```diff
@@ -696,3 +696,132 @@ func isZero(v reflect.Value) bool {
...
+		}
+	case yaml_DOCUMENT_END_EVENT:
+		if e.implicit {
+			str = "-DOC"
+		} else {
+			str = "-DOC---"
```
perlpunk:
```suggestion
			str = "-DOC ..."
```

yaml.go
pantoniou: Indeed, merged (in my fork that is)

perlpunk:
This would be really nice to have, as we could integrate go-yaml into the test matrix: https://matrix.yaml.info/

yaml.go
```diff
@@ -696,3 +696,132 @@ func isZero(v reflect.Value) bool {
...
+			str = "+DOC ---"
+		}
+	case yaml_DOCUMENT_END_EVENT:
+		if e.implicit {
+			str = "-DOC"
+		} else {
+			str = "-DOC ..."
+		}
+	case yaml_MAPPING_START_EVENT:
+		str = "+MAP"
```
perlpunk, line 776: can go-yaml differentiate between flow and block mappings here? (same question for sequences below)

inteon:
@pantoniou @perlpunk Great PR! In https://github.com/amurant/go-yaml/tree/v3_next I further explored this idea and made it into a working test. Feel free to continue on that and leave a PR on that repo.

not sure if we will be able to merge it back into this repo, since there seems to be no merge-activity here anymore ...

pantoniou:
@inteon Interesting. I will take a look.

FWIW I've started work in a complete golang binding of my library libfyaml https://github.com/pantoniou/go-fyaml

Note that it's not nearly complete yet, but I intent to get it to a point where it's complete enough to work for most usage cases. I would appreciate it if you could take a look.

PR #819, opened 2022-01-29
Cherrypick to v2: Use context line for reporting

artemmikhalitsin:
kubernetes-sigs/yaml uses v2 of go-yaml, and this fix is necessary for correctly identifying where errors are in large yaml files, so it would be great to have this added to v2.

For example, I'd like to have this for debugging charts in helm: https://github.com/helm/helm/pull/8699


max-rocket-internet:
Any update on this? Could we get a review @niemeyer?

PR #820, opened 2022-01-30
fix: Keep quotes on very big binary and hexadecimal integer strings

artemmikhalitsin:
reported in a related issue https://github.com/helm/helm/issues/4262

Binary and hexadecimal integer strings which are larger than the max unit64 number are currently being marshalled using yaml_PLAIN_SCALAR_STYLE. This fix is so they are marshalled with quotes using yaml_LITERAL_SCALAR_STYLE.

Note that the decimal (and probably octal) strings are already marshalled using yaml_LITERAL_SCALAR_STYLE because they happen to match the yamlStyleFloat regexp, and are successfully parsed as float.


almson:
Maybe add some tests to make sure octals, numbers, booleans, and many other things keep their quotes in the future?

artemmikhalitsin:
Thanks for the pointer, I added some

Issue #818, opened 2022-01-26
yaml to struct - merge behavior : deep vs shallow merge 


kansalarchit:
It seems like anchor merge behaves differently depending on the golang struct provided.

With a simple golang interface{} object, the conversion results in shallow merge. e.g.:

```
top:
  config: &config
    key1: value1
    map1: 
      mapkey1: value1
      mapkey2: value2

  otherconfig:
    <<: *config
    map1:
      mapkey1: some-other-value
```

results in (Shallow Merge)
```
top:
  config: &config
    key1: value1
    map1: 
      mapkey1: value1
      mapkey2: value2

  otherconfig:
    key1: value1
    map1:
      mapkey1: some-other-value
```

whereas when we provide a specific golang struct, such as

```
type top struct  {
  type config struct {
    key1 string,
    type map1 struct {
      mapkey1 string,
      mapkey2 string
      ...
    }
    ...
  }
  ...
}
```

For the same input, we get: (Deep Merge)
```
top:
  config: &config
    key1: value1
    map1: 
      mapkey1: value1
      mapkey2: value2

  otherconfig:
    key1: value1
    map1:
      mapkey1: some-other-value
      mapkey2: value2
```

This looks like inconsistent behavior and based on https://yaml.org/type/merge.html I believe we should always prefer shallow merge.

Can a flag be added to the parser so that we can choose to ignore deep merge even when parser knows the resulting struct ?

niemeyer:
This and a few more complex cases should have been fixed by 539c8e7.

Please let me know if you see any further issues.

PR #826, opened 2022-02-28
v3: Enabled shallow merge when unmarshal to struct.

mattaldenshaw:
This creates a flag called "shallowMerge" in the decoder to force the merge to be shallow when unmarshalling from yaml to struct. It will fix issue #818 @niemeyer

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -315,11 +315,12 @@ type decoder struct {
 	stringMapType  reflect.Type
 	generalMapType reflect.Type
 
-	knownFields bool
-	uniqueKeys  bool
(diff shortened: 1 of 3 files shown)


niemeyer:
As we discussed on the call, the proper solution to this problem is to make the decoder behave correctly per the specification and the other popular parsers, all the time. Decoding a document containing a merge should be no different than decoding a simpler document that results from the merge operations themselves, for all purposes.

I believe I've addressed that in 539c8e7.

Please let me know if you see any issues there.

PR #828, opened 2022-03-05
Replace initialization code for decode ifaceType

ZauberNerd:
Tinygo currently does not support reflect.Elem() on a map value.
In order to be able to compile programs that use yaml.v3 with tinygo
(for example to compile to wasm), we should replace the variable
initialization of ifaceType with something that tinygo understands.

The tinygo reflect `Type.Elem()` currently does not support the `map` type: https://github.com/tinygo-org/tinygo/blob/3883550c446fc8b14115366284316290f3d3c4b1/src/reflect/type.go#L348-L365
And also the interpreter of tinygo does not have support for it: https://github.com/tinygo-org/tinygo/blob/3883550c446fc8b14115366284316290f3d3c4b1/interp/interpreter.go#L363-L370

But there are two PRs open with major refactorings for both of these systems, so maybe in the future it will have support for it:
* https://github.com/tinygo-org/tinygo/pull/2640
* https://github.com/tinygo-org/tinygo/pull/2655

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -323,11 +323,12 @@ type decoder struct {
 }
 
 var (
+	sliceOfEmptyInterface []interface{}
 	nodeType       = reflect.TypeOf(Node{})
 	durationType   = reflect.TypeOf(time.Duration(0))
 	stringMapType  = reflect.TypeOf(map[string]interface{}{})
 	generalMapType = reflect.TypeOf(map[interface{}]interface{}{})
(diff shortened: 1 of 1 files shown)


ZauberNerd:
After having thought about this for a while, I'll close this PR. It doesn't fix a bug in go-yaml and it does not improve its code.
The correct way forward is to wait for the above linked PRs in tinygo to land or improve upon them myself.

Issue #833, opened 2022-03-18
Panic when parsing "! !00 \xf6" string


prymitive:
Found via fuzzing tests added to Go 1.18

```go
package main

import (
	"fmt"

	"gopkg.in/yaml.v3"
)

func main() {
	defer func() {
		if r := recover(); r != nil {
			fmt.Printf("panic: %s", r)
		}
	}()

	var node yaml.Node
	yaml.Unmarshal([]byte(string("! !00 \xf6")), &node)
}
```

https://go.dev/play/p/28Dlf2ncCct

Zenithar:
Same with `string("! !!0 \xf7")`

tcarreira:
I think this issue is fixed already.
At least the [example](https://go.dev/play/p/28Dlf2ncCct) you gave passes fine with no panic.

prymitive:
#666 fixed it

dolmen:
@prymitive Could you submit your fuzzing code?

Issue #666, opened 2020-10-21
v3: panic "attempted to parse unknown event (please report): none"


bradleyjkemp:
Hi folks 👋🏻 Found this panic (along with #665) while fuzzing my own project.

Minimal example of the panic (https://play.golang.org/p/gLM_eHzcrgz):
```go
package main

import (
	"gopkg.in/yaml.v3"
)

func main() {
	var t interface{}
	yaml.Unmarshal([]byte("0: [:!00 \xef"), &t)
}
```

Output:
```
panic: internal error: attempted to parse unknown event (please report): none [recovered]
	panic: internal error: attempted to parse unknown event (please report): none

goroutine 1 [running]:
gopkg.in/yaml%2ev3.handleErr(0xc000043f60)
	/tmp/gopath818249317/pkg/mod/gopkg.in/yaml.v3@v3.0.0-20200615113413-eeeca48fe776/yaml.go:294 +0x85
panic(0x50f120, 0xc000010330)
	/usr/local/go-faketime/src/runtime/panic.go:969 +0x1b9
gopkg.in/yaml%2ev3.(*parser).parse(0xc000036c00, 0x0)
	/tmp/gopath818249317/pkg/mod/gopkg.in/yaml.v3@v3.0.0-20200615113413-eeeca48fe776/decode.go:163 +0x277
gopkg.in/yaml%2ev3.(*parser).parseChild(...)
	/tmp/gopath818249317/pkg/mod/gopkg.in/yaml.v3@v3.0.0-20200615113413-eeeca48fe776/decode.go:194
gopkg.in/yaml%2ev3.(*parser).sequence(0xc000036c00, 0xc00000e007)
	/tmp/gopath818249317/pkg/mod/gopkg.in/yaml.v3@v3.0.0-20200615113413-eeeca48fe776/decode.go:259 +0xff
gopkg.in/yaml%2ev3.(*parser).parse(0xc000036c00, 0x0)
	/tmp/gopath818249317/pkg/mod/gopkg.in/yaml.v3@v3.0.0-20200615113413-eeeca48fe776/decode.go:154 +0xe7
gopkg.in/yaml%2ev3.(*parser).parseChild(0xc000036c00, 0xc00007e3c0, 0xc00007e460)
	/tmp/gopath818249317/pkg/mod/gopkg.in/yaml.v3@v3.0.0-20200615113413-eeeca48fe776/decode.go:194 +0x2f
[...]

PR #850, opened 2022-05-10
fix(decoder): panic raised on fuzzer inputs.

Zenithar:
Handle yaml_NO_EVENT event type to raise a parsing
error vs panic.

ref #833

Signed-off-by: Thibault Normand <me@zenithar.org>


Zenithar:
Solved by https://github.com/go-yaml/yaml/issues/666 - https://github.com/go-yaml/yaml/commit/8f96da9f5d5eff988554c1aae1784627c4bf6754

PR #856, opened 2022-05-23
fix: re-add option to switch off unique mapping keys enforcement

jlourenc:
This PR partially reverts https://github.com/go-yaml/yaml/commit/309c9d15755818936fcb0f5c171692693a47efe2 while keeping unique mapping keys enforcement as default behaviour for `Unmarshal`.

The rationale is that libraries like `ghodss/yaml` relies on `uniqueKeys` being set to `false`. Giving the possibility to do so would allow the upgrade of `go-yaml` to `v3` without introducing breaking changes.


jlourenc:
`ghodss/yaml` draft fix using this changeset: https://github.com/ghodss/yaml/pull/79

jlourenc:
@niemeyer could I get some eyes on this ? That'd be much appreciated 🙇🏼

PR #640, opened 2020-07-31
Discount space taken by sequence indicator

abursavich:
Fixes #633 and #643

PR #864, opened 2022-06-10
fix: scalar indent indicator [#643]

### LINKED ISSUE #643: v3: Invalid output (bad indentation) when an indentation indicator/hint is used in a scalar block
The indentation change in https://github.com/go-yaml/yaml/commit/a5ece683394c3b88d90572e44d36c93aea492c2c is causing  invalid output when an indentation indicator/hint is used in a scalar block: https://play.golang.org/p/tbWDmsaqANV

for:
```go
yaml.Marshal([]string{" hello\nworld"})
```
expecting:
```yaml
- |4-
     hello
    world
```
got:
```yaml
- |4-
   hello
  world
```
which is invalid and cannot be parsed.

This relates to #633 and I had commented about it there, but invalid output is more severe than unpleasant indentation... so I figured it deserved a separate issue.

hedhyw:
Fixes #643

The library can produce invalid YAML (`yaml.Marshal([]string{" hello\nworld"})`):
```yaml
- |4-
   hello
  world
```

The indention is `2`, but the indicator shows `4-`.

This PR fixes this behaviour.

### DIFF
--- a/emitterc.go
+++ b/emitterc.go
@@ -162,10 +162,9 @@ func yaml_emitter_emit(emitter *yaml_emitter_t, event *yaml_event_t) bool {
 // Check if we need to accumulate more events before emitting.
 //
 // We accumulate extra
-//  - 1 event for DOCUMENT-START
-//  - 2 events for SEQUENCE-START
-//  - 3 events for MAPPING-START
-//
+//   - 1 event for DOCUMENT-START
+//   - 2 events for SEQUENCE-START
+//   - 3 events for MAPPING-START
 func yaml_emitter_need_more_events(emitter *yaml_emitter_t) bool {
 	if emitter.events_head == len(emitter.events) {
 		return true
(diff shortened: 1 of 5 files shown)


emitterc.go
```diff
@@ -1847,7 +1847,7 @@ func yaml_emitter_write_double_quoted_scalar(emitter *yaml_emitter_t, value []by
 
 func yaml_emitter_write_block_scalar_hints(emitter *yaml_emitter_t, value []byte) bool {
 	if is_space(value, 0) || is_break(value, 0) {
-		indent_hint := []byte{'0' + byte(emitter.best_indent)}
+		indent_hint := []byte{'0' + byte(emitter.indent)}
```
emanlodovice: Thanks for the fix, it seems like this change does not work in some cases. For example

```
package main

import (
	"fmt"

	"gopkg.in/yaml.v3"
)

type T struct {
	F      int `yaml:"a,omitempty"`
	B      string
	Values []T
}

func main() {
	data := &T{B: "test1"}

	withNewLine := `
line 1
  line 2

  line 3

`
	data.Values = append(data.Values, T{B: withNewLine})
	val, _ := yaml.Marshal(data)
	fmt.Println(string(val))
}
```
produces 
```
b: test1
values:
    - b: |8+
        line 1
          line 2

          line 3

      values: []
```
which is invalid.

I found that the changes:
https://github.com/go-yaml/yaml/blob/6367991e90d61590c8621de8aa4e562966d0822a/emitterc.go#L727-L730
https://github.com/go-yaml/yaml/blob/6367991e90d61590c8621de8aa4e562966d0822a/emitterc.go#L237-L239

from the PR https://github.com/go-yaml/yaml/pull/640 for the issue that you linked worked.
Do you mind adjusting your PR with these changes? Thank you.

emanlodovice (changes requested):

emitterc.go
hedhyw: Hi, sure

hedhyw: @emanlodovice I applied those changes and added more test cases

emanlodovice (approved):
LGTM

emanlodovice:
Approved the PR but I am not a maintainer 😅

emanlodovice:
@niemeyer please take a look when you have time. Thank you

Issue #797, opened 2021-10-28
Error on optional string escapes, even though valid in JSON


chaimleib:
**Describe the bug**
According to the [JSON spec](https://datatracker.ietf.org/doc/html/rfc8259#page-8) (YAML is a superset of JSON), [YAML 1.2.2](https://spec.yaml.io/main/spec/1.2.2/#57-escaped-characters), and [YAML 1.3.0 (draft 2022-01-19)](https://spec.yaml.io/main/spec/1.3.0/#escaped-characters), `"\/"` is valid string syntax, equivalent to `"/"`.

However, when `"\/"` exists on the input to `go-yaml/yaml`, the escaped character is not recognized.

Versions of go-yaml: v2.4.0, both v2 and v3 branches
Operating system: mac, Go Playground

**Example code**
https://play.golang.org/p/UbP6fq9WPbd
```go
package main

import (
	"fmt"
	"strings"

	v2 "gopkg.in/yaml.v2"
	v3 "gopkg.in/yaml.v3"
)

var s = `"\/"`

func main() {
	var out interface{}
	if err := v2.Unmarshal([]byte(s), &out); err != nil {
		fmt.Printf("v2: error: %v\n", err)
	} else {
		fmt.Printf("v2: %#v\n", out)
	}

	reader := strings.NewReader(s)
	var node v3.Node
	decoder := v3.NewDecoder(reader)
	if err := decoder.Decode(&node); err != nil {
		fmt.Printf("v3: error: %v\n", err)
	} else {
		fmt.Printf("v3: %#v\n", node.Content[0].Value)
	}
}
```

**Actual behavior**

```
v2: error: yaml: found unknown escape character
v3: error: yaml: found unknown escape character
```

**Expected behavior**

```
v2: "/"
v3: "/"
```

perlpunk:
note: YAML 1.3.0 doesn't exist yet, it's work in progress and shouldn't be linked to

rasentry:
@perlpunk this means this problem will not be fixed?

perlpunk:
@rasentry I don't understand your question. *What* will not be fixed?

I'm a member of the YAML core team and only wanted to note  that YAML version 1.3 does not exist yet, and noone should link to it for reference.

I will ask the members of the team maintaining the site again to make this more clear that this is a draft.

I'm not a developer of go-yaml.
go-yaml should accept `"\/"` if it implements YAML 1.2.
If it implements YAML 1.1 only, it behaves correctly.

[...]

PR #862, opened 2022-06-04
fix: Support escaped slash in strings

hairyhenderson:
Fixes #797

The JSON spec lists the `/` character as optionally-escapable, and so YAML v1.2.0 added support for the sequence `\/` to be unescaped to `/`.

This PR adds the minimum necessary support for this, and should solve the issues that a few people have already hit, as documented in #797.

Thanks!

Signed-off-by: Dave Henderson <dhenderson@gmail.com>

PR #884, opened 2022-08-08
V2: fix: Support escaped slash in strings

chaimleib:
> Fixes https://github.com/go-yaml/yaml/issues/797
> 
> The JSON spec lists the / character as optionally-escapable, and so YAML v1.2.0 added support for the sequence \/ to be unescaped to /.
> 
> This PR adds the minimum necessary support for this, and should solve the issues that a few people have already hit, as documented in https://github.com/go-yaml/yaml/issues/797.

This is a git cherry-pick of @hairyhenderson's PR #862, but targeted for the V2 branch. See his PR for the V3 fix.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -734,6 +734,15 @@ b:
 				"a": "a",
 			}},
 	},
+
+	// Issue #797 - escaped slash
+	{
+		"a: \"\\/\"\nb: /",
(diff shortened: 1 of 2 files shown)


hairyhenderson:
thanks for this @chaimleib 🙇‍♂️

PR #894, opened 2022-08-27
scannerc: Added support for CRLF edge case

### LINKED ISSUE #893: Decoding documents with CRLF line endings parses an extra new line
Decoding a document like this:
```yaml
# comment
a:
```
And then encoding again leads to a result like this:
```yaml
# comment

a:
```

The problem is that there is an edge case where the scanner peeks ahead and finds an `LF` byte, and thinking it's in a valid line break position it appends a `\n` to the text. However, this ignores the possibility that this is actually the `LF` of a complete `CRLF`.

braydonk:
Closes #893 

There is an edge case where the scanner peeks ahead and finds an LF byte,
and thinking it's in a valid line break position it appends a \n to the
text. However, this ignores the possibility that this is actually the LF
of a complete CRLF.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -25,8 +25,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/braydonk/yaml"
 	. "gopkg.in/check.v1"
-	"gopkg.in/yaml.v3"
 )
 
 var unmarshalIntTest = 123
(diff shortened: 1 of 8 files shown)


braydonk:
I'm not 100% sure where to put a test for this particular case, I will revisit next time I'm available.

braydonk:
Ended up merging this into my own fork and did some rebasing stuff so can't merge this particular PR here, but happy to re-make it if the contribution is wanted.

PR #905, opened 2022-09-23
Allow question marks in plain scalars in flow collections

chris-olszewski:
Allow question marks in plain scalars in flow collections

According to the YAML 1.1 spec question marks can be allowed in plain scalars that appear in flow collections.

This PR copies the changes from https://github.com/yaml/libyaml/pull/105 to allow for question marks to appear in the middle of plain flow scalars.
As the original PR states this doesn't allow for `?foo` yet even though that is valid according to the YAML spec.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -802,6 +802,15 @@ var unmarshalTests = []struct {
 			"c": []interface{}{"d", "e"},
 		},
 	},
+
+	// Question mark in plain scalar flow collection
+	{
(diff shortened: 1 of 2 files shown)


weyert:
I think @andig or @niemeyer might be able to review this PR?

andig:
I have no idea, sorry

tjx666:
@niemeyer

PR #913, opened 2022-10-12
Make possible to implement Unmarshaler for map types.

alex-ac:
The more-or-less canonical way to implement set data structure in go is by using `map[key]struc{}` type. That type can implement `Marshaler`/`Unmarshaler` interfaces to define custom marshaling (into a sequence of items, for example). Unfortunatelly, current code does not try to check if map type implements `Unmarshaler` interface because map is not addressable.

Simple snippet to check it:

```
package main
import "gopkg.in/yaml.v3"

type stringSet map[string]struct{}

// Marshaler implementation marshals this type as sequence. This part works flawlessly
func (s stringSet) MarshalYAML() (interface{}, error) {
  values := make([]string, 0, len(s))
  for v, _ := range s {
    values = append(values, v)
  }
  return values
}

// Unmarshaler implementation is not called during unmarshaling.
func (s stringSet) UnmarshalYAML(value *yaml.Node) error {
  values  := make([]string, 0, len(value.Content))
  if err := value.Decode(&values); err != nil {
    return err
  }
  newSet := make(map[string]struct{}, 0, len(values))
  for _, v := range values {
    newSet[v] = struct{}{}
  }
  s = newSet
  return nil
}

func main() {
  set := stringSet{}
  data := "[a, b, c]"
  if err := yaml.Unmarshal([]byte(data), set); err != nil {
    // Currently returns: yaml: unmarshal errors:
    //      line 1: cannot unmarshal !!seq into main.stringSet
    panic(err)
  }
}
```

This pull-request:
1. Adds a very specific test which is almost the same as snippet before. If you run tests on the first commit of the PR, that tests fails.
2. Slightly changes logic of `decoder.prepare` function. Now that function tries to cast a reflected value to `Unmarshaler` interface even if it is not addressable. That allows calling UnmarshalYAML on map types.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -1708,7 +1708,7 @@ type stringSetUnmarshaler map[string]struct{}
 func (s stringSetUnmarshaler) UnmarshalYAML(value *yaml.Node) error {
 	values := make([]string, 0, len(value.Content))
 
-	if err := value.Decode(values); err != nil {
+	if err := value.Decode(&values); err != nil {
 		return err
 	}
 
@@ -1725,10 +1725,8 @@ func (s stringSetUnmarshaler) UnmarshalYAML(value *yaml.Node) error {
 
 func (s *S) TestStringSetUnmarshaler(c *C) {
 	cases := []string{
-		"a:\nb:\nc:",
 		"- a\n- b\n- c",
 		"[a, b, c]",
-		"{a: null, b: null, c: null}",
 	}
 
 	for _, data := range cases {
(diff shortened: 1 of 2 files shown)


decode_test.go
```diff
@@ -947,7 +947,7 @@ var unmarshalErrorTests = []struct {
 	{"%TAG !%79! tag:yaml.org,2002:\n---\nv: !%79!int '1'", "yaml: did not find expected whitespace"},
 	{"a:\n  1:\nb\n  2:", ".*could not find expected ':'"},
 	{"a: 1\nb: 2\nc 2\nd: 3\n", "^yaml: line 3: could not find expected ':'$"},
-	{"#\n-\n{", "yaml: line 3: could not find expected ':'"}, // Issue #665
```
alex-ac, line 950: gofmt did a little extra changes on this file.

PR #921, opened 2022-11-12
Support scanning unicode characters in anchor names

### LINKED ISSUE #920: Anchors only support a limited subset of characters
The YAML spec allows for anchors to contain any non-whitespace printable character except for `[]{},`.
https://yaml.org/spec/1.2.2/#692-node-anchors

The implementation currently only supports ascii alphanumerics `a-zA-Z0-9_-`.

michaelruigrok:
Fixes #920

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -494,6 +494,38 @@ var unmarshalTests = []struct {
 	}, {
 		"a: &a [1, 2]\nb: *a",
 		&struct{ B []int }{[]int{1, 2}},
+	}, {
+		// >= 0x21
+		"a: &! [1, 2]\nb: *!",
+		&struct{ B []int }{[]int{1, 2}},
+	}, {
+		// <= 0x7E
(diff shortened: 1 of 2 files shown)


michaelruigrok:
You may wish to repurpose the check for valid anchor name characters into `is_printable()`. They currently have different purposes, with mine checking if a character is within's YAML's official list of printable characters, and the other deciding itself what is a practical character set to encode on output. But one might believe these character sets should be aligned.

One similar issue I believe to be larger is the treatment of whitespace characters, for which I have made an issue: #922. My implementation here uses `isblankz()`, which treats non-ascii line breaks as line breaks, contrary to the YAML 1.2 spec. I have included these characters when checking for allowed printable characters, but then defer to `blankz()` to reject any characters that it deems as whitespace, to maintain consistency with the rest of the codebase. If `isblankz` is updated to treat these characters as not whitespace, the anchor scanner will accept them as part of the anchor name.

PR #935, opened 2023-01-30
Feature: Initial support of custom tag handler for v3 for scalar values

lelvisl:
Hi!

In this PR I want add initial support of custom tag handlers for scalar values

example code

```go
package main

import (
	"fmt"
	"gopkg.in/yaml.v3"
	"log"
	"os"
	"strings"
)

var data = `
a: !envstring SECRET
b: 
  c: 85
  d: [3, 4]
`

func main() {
	os.Setenv("SECRET", "pass")
	fmt.Println("hi")
	var foo struct {
		A string
		B interface{}
	}
	dec := yaml.NewDecoder(strings.NewReader(data))
	dec.SetTagHandler("!envstring", EnvTagHandler{}.Handle)
	err := dec.Decode(&foo)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(foo)
	fmt.Println(foo.A)
}

type EnvTagHandler struct{}

func (e EnvTagHandler) Handle(str string) (any, error) {
	return os.Getenv(str), nil
}
```

In future all default tags can be implemented as tag handlers.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -326,7 +326,8 @@ type decoder struct {
 
 	mergedFields map[interface{}]bool
 
-	tagHandlers map[string]TagHandler
+	tagResolvers map[string]TagResolver
+	tagReplacers map[string]TagReplacer
 }
 
 var (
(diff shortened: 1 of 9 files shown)


lelvisl:
@niemeyer please look at this PR. If you basically agree, i will add tests.

PR #936, opened 2023-01-31
Add String() methods to Kind and Style types

moskyb:
We found that when debugging, it's difficult to know what Kind/Style different nodes are (see: https://github.com/buildkite/agent/pull/1930#discussion_r1090068650) without looking at the source code and going through the iotas.

This PR adds `String()` methods to both the kind and the type structs, such that they fulfil the `Stringer` interface, and so can be printed in a human-readable way using the `%v` Printf verb.

### DIFF
--- a/yaml.go
+++ b/yaml.go
@@ -330,6 +330,24 @@ const (
 	AliasNode
 )
 
+// String returns the string representation of the node kind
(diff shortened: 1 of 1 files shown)


aathan:
Or just `%+v` ?

PR #943, opened 2023-02-28
Set top-level minimal permissions on Github workflows

### LINKED ISSUE #941: CI: Set minimum permissions on GitHub Workflow
Hi!

I'm Diogo and I work on Google's Open Source Security Team([GOSST](https://github.com/diogoteles08#about-gosst-ghost)) in cooperation with the Open Source Security Foundation ([OpenSSF](https://openssf.org/)). My core job is to suggest and implement security changes on widely used open source projects 😊

I'm here to suggest the definition of minimal permissions on your workflow, as it would harden your security agains supply-chain attacks. I see that you have only one workflow, the [go.yml](https://github.com/go-yaml/yaml/blob/v3/.github/workflows/go.yaml), but it does not specify the permissions for its jobs, letting their privileges determined by GitHub's defaults. By defining minimal permissions you would be secured against erroneous or malicious actions from external jobs you call from your workflow. It's specially important for the case they get compromised, for example.

Setting minimum permissions for workflows is recommended by [GitHub itself](https://docs.github.com/en/actions/security-guides/automatic-token-authentication) and also by other security tools, such as [Scorecards](https://github.com/ossf/scorecard) and [StepSecurity](https://github.com/step-security).

As the changes would be very simple, I'll take the liberty to raise a PR and ease your evaluation of the changes.

diogoteles08:
It secures the repo against erroneous or malicious actions from external jobs called from the workflow. It's specially important for the case they get compromised, for example.

Closes #941

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -1,6 +1,8 @@
 ---
 name: Go
 on: [push, pull_request]
+permissions:
+  contents: read
 jobs:
   test:
     name: Test


diogoteles08:
Closing due to inactivity, as detailed on the original issue #941

PR #945, opened 2023-03-03
support merge of map sequence alias

davidmdm:
the [merge operator](https://yaml.org/type/merge.html) is defined as such that a merge operation can work with a map or sequence of maps. 

IE both of the following examples are legal:

```yaml
node:
  <<: key: value

# same as:
node:
  key: value  
```

and:
```yaml
node:
  <<:
    - k1: v1
    - k2: v2

# same as
node:
  k1: v1
  k2: v2
```

Today we are able to expand an alias node for a merge operation using `go-yaml` which supports the first case:

```yaml
base: &base
  key: value

node:
  <<: *base
```

However `go-yaml` does not support map sequence aliases:
```yaml
base: &base
  - k1: v1
  - k2: v2

node:
  <<: *base
```

When unmarshaling this we get the following error: `map merge requires map or sequence of maps as the value`.
Which is technically wrong because we are providing a sequence of maps.

This PR fixes this case.


davidmdm:
@niemeyer Please take a look when you have a chance!

vallieres (approved):
LGTM

mfridman:
This is great, would love to see this merged in eventually.

PR #960, opened 2023-05-03
Fix new line on encoding folded style

### LINKED ISSUE #804: v3: FoldedStyle yaml.Node results in unexpected newlines
When Marshalling/Encoding a scalar YAML node with the folded multi-line style I see an unexpected new line in the encoded YAML.

I've added some test cases in a [branch](https://github.com/go-yaml/yaml/compare/v3...janakerman:jan-folded-newline-issue) in my fork.

A `yaml.Node` with `style.Folded` and the following `Value`:
```
foo
    bar
```

Is marshalled with a newline between foo and bar:
```
>-
    foo

        bar
```

I'm expecting it to be marshalled as follows:
```
>-
    foo
        bar
```

Reading the [spec](https://yaml.org/spec/1.2-old/spec.html#style/block/folded) I believe my expectations are correct.

> Lines starting with white space characters (more-indented lines) are not folded.

The snippet `Example 8.10. Folded Lines` shows an indented line and it does not include the additional newline.

The impact of this is that marshalling unmarshalling a YAML string into a YAML node and back into a string again does not result in the same string value within the node.

zecke:
When using:

```
 >-
   someline
     more

The encoder would emit:

 >-
   someline

     more
```

The issue was introduced in the original translation from C to Go. In the C code the MOVE macro would advance the start pointer of the string. In the Go case "k = 0" points to the start of the string where in the C case it would be the read pointer. Fix it.

Also address bound checks to not go beyond the string (I assume in the C code there is an explicit NUL or at least some NUL).

Fixes: #804

### DIFF
--- a/emitterc.go
+++ b/emitterc.go
@@ -1943,11 +1943,11 @@ func yaml_emitter_write_folded_scalar(emitter *yaml_emitter_t, value []byte) boo
 	for i := 0; i < len(value); {
 		if is_break(value, i) {
 			if !breaks && !leading_spaces && value[i] == '\n' {
-				k := 0
-				for is_break(value, k) {
+				k := i
+				for k < len(value) && is_break(value, k) {
 					k += width(value[k])
 				}
-				if !is_blankz(value, k) {
+				if k < len(value) && !is_blankz(value, k) {
 					if !put_break(emitter) {
 						return false
 					}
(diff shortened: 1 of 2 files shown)


zecke:
@niemeyer Any chance you can have a look?

dolmen:
@zecke In #1028 I'm proposing to add a fuzzer.

PR #961, opened 2023-05-07
Unmarshal scalar binary data into `[]byte` 

### LINKED ISSUE #231: Encode []byte as base64 string (to be consistent with stdlib JSON)
The standard library JSON module automatically converts any `[]byte` struct members to base64 encoded strings when marshalling (see https://golang.org/pkg/encoding/json/#Marshal). The YAML library does not, and instead outputs a `[]byte` as an array of bytes, like so:

```YAML
bytes:
  - 54
  - 32
```

I think it would be nice for go-yaml to be consistent with the go standard library behaviour for JSON.

danielorbach:
The YAML spec supports binary data (http://yaml.org/type/binary.html), however this package decodes and encodes nodes with this tag as strings.
This conflicts with the standard Go type `[]byte` - which is used interchangeably with `string`, especially in the context of binary blobs.

Closes #231

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -615,6 +615,13 @@ func (d *decoder) scalar(n *Node, out reflect.Value) bool {
 		}
 		out.SetString(n.Value)
 		return true
+	case reflect.Slice:
+		if out.Type().Elem().Kind() == reflect.Uint8 {
+			if tag == binaryTag {
+				out.SetBytes([]byte(resolved.(string)))
+				return true
+			}
+		}
 	case reflect.Interface:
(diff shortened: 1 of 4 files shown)


danielorbach:
Hey @niemeyer !

This PR continues on your work on bc27649cd5454055cf20fdb9ef556c214d3f9aa0 a while back (like 9 years ago) .
I'd really appreciate your review on this changeset.

Getting it into the upstream would really help my company adopt YAML as our standard configuration format.

P.S. tests pass on my machine (although I've had to run `go mod tidy` after a fresh clone)

PR #962, opened 2023-05-22
Add time formats to allowedTimestampFormats

facundopoblete:
PR to extend the allowedTimestampFormats list. For example, the time format without a timezone was only valid if it was separated by a space.

Related issue: #192


facundopoblete:
Hi @niemeyer / @rogpeppe! I would really appreciate your feedback on this change/quick fix. For some reason, when the fix for #192 was introduced, the only supported format for datetime without a timezone was the one with a space as a separator.

Issue #918, opened 2022-11-02
Error while merging anchors


AbrahamTewa:
Hi !

I have an error when I try to unmarshall YAML with merging anchors:

```go
package main

import (
	"log"

	yaml "gopkg.in/yaml.v3"
)

var data = `
anchorA: &anchorA
  A: 1

  anchorB: &anchorB
  B: 2

testFinal:
  <<: *anchorA
  <<: *anchorB
`

func main() {
	m := make(map[interface{}]interface{})

	err := yaml.Unmarshal([]byte(data), &m)
	log.Fatalf("error: %v", err)
}
```

I get an error:
```
error: yaml: unmarshal errors:
  line 10: mapping key "<<" already defined at line 9
```

perlpunk:
try this:
```
testFinal:
  <<: [*anchorA, *anchorB]
```

AbrahamTewa:
This syntax seems to work (no errors).

AbrahamTewa:
Hi !
Any update on this issue ?

Thanks !

perlpunk:
I wouldn't expect an update. My suggestion is the official way to merge multiple anchors, see https://yaml.org/type/merge.html

go-yaml's behaviour is not a bug. You better rewrite your YAML.

PR #976, opened 2023-07-21
Multiple merge

### LINKED ISSUE #966: [BUG]: Aliases does not unmarshal (mapping key "<<" already defined ...)
I am currently using the v3 of go-yaml , the problem is that unmarshalling multiple alias throws `mapping key "<<" already defined at line 10`. Below is the example code and yaml file

```main.go
package main

import (
	"fmt"
	"log"
	"os"

	"gopkg.in/yaml.v3"
)

type Foo struct {
	Foo map[string]interface{} `yaml:"foo"`
}

func main() {

	file, err := os.Open("./test.yaml")

	if err != nil {
		log.Fatal(err)
	}

	x, _ := file.Stat()

	content := make([]byte, x.Size())

	_, err = file.Read(content)

	if err != nil {
		log.Fatal(err)
	}

	c := Foo{}

	err = yaml.Unmarshal(content, &c)

	if err != nil {
		log.Fatal(err)
	}

	fmt.Println(c)
}
```

```test.yaml
reuse:
  - &reuse
    bar: "baz"

reuse2:
  - &reuse2
    baz: "foobar"

foo:
  <<: *reuse
  <<: *reuse2

```

moskyb:
Fixes #966

This PR adds the capability to decode multiple merge nodes on a given document - eg, the document
```Yaml
one: &one
  a: 1
  b: 2

two: &two
  b: 3
  c: 4

merged:
  <<: *one
  <<: *two
```

will now decode to 
```Go
map[string]any{
  "a": 1
  "b": 3
  "c": 4
}
```

where previously, it would have returned an error saying that the `merged` mapping had a duplicate key `<<`.

While this behaviour isn't in the YAML spec (the blessed way to do multiple merges is `<<: [*one, *two]`), it's also not disallowed by it, and many yaml libraries for other languages - notably Python's [pyyaml](https://github.com/yaml/pyyaml) and Ruby's [psych](https://github.com/ruby/psych), which is backed by [libyaml](https://github.com/yaml/libyaml) - do support this behaviour. 

Multiple issues (#966, #918) on this have been raised on this repo before, which leads me to believe that using multiple merge keys is a relatively common expectation to have for a yaml library.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -770,6 +770,11 @@ func (d *decoder) mapping(n *Node, out reflect.Value) (good bool) {
 		nerrs := len(d.terrors)
 		for i := 0; i < l; i += 2 {
 			ni := n.Content[i]
+			if isMerge(ni) {
+				// Merge keys ("<<") are discarded during processing, so don't need to be unique.
+				continue
+			}
+
 			for j := i + 2; j < l; j += 2 {
 				nj := n.Content[j]
 				if ni.Kind == nj.Kind && ni.Value == nj.Value {
@@ -816,7 +821,7 @@ func (d *decoder) mapping(n *Node, out reflect.Value) (good bool) {
 	mergedFields := d.mergedFields
 	d.mergedFields = nil
 
-	var mergeNode *Node
+	mergeNodes := []*Node{}
 
 	mapIsNew := false
 	if out.IsNil() {
@@ -825,7 +830,7 @@ func (d *decoder) mapping(n *Node, out reflect.Value) (good bool) {
 	}
 	for i := 0; i < l; i += 2 {
 		if isMerge(n.Content[i]) {
-			mergeNode = n.Content[i+1]
+			mergeNodes = append(mergeNodes, n.Content[i+1])
 			continue
 		}
 		k := reflect.New(kt).Elem()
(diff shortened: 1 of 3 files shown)


decode.go
```diff
@@ -772,7 +772,8 @@ func (d *decoder) mapping(n *Node, out reflect.Value) (good bool) {
 			ni := n.Content[i]
 			for j := i + 2; j < l; j += 2 {
 				nj := n.Content[j]
-				if ni.Kind == nj.Kind && ni.Value == nj.Value {
+				// Key collisions are allowed if both of the keys are merges - each merge will be applied in order, leading to no collisions in the final decoded document
+				if ni.Kind == nj.Kind && ni.Value == nj.Value && !isMerge(nj) {
```
pda: We could probably hoist this above the `for j := …` loop, and say:

```go
if isMerge(ni) {
  // Merge keys ("<<") are discarded during processing, so don't need to be unique.
  continue
}
```

pda (commented):
It would be good to match how other applications of libyaml¹ handle https://yaml.org/type/merge.html by supporting multiple merge keys like this 👍🏼

It's true that the Merge Key spec shows passing an array of anchors to achieve this result, rather than multiple merge keys in a single mapping, but the fact that it's supported in other popular languages, and therefore used out in the wild², makes this an interoperability issue.

¹ While go-yaml isn't strictly an _application of libyaml_, it is a Go port of libyaml _plus_ the Go application/integration thereof, so I think it's the same category.

² I work with @moskyb and we have examples of multiple customers using the multiple-Merge-Key style.

derbylock:
Can we release it?

PR #979, opened 2023-08-09
better detection of pointer receivers

aathan:
I did not attempt to fully grok the yaml marshaler implementation but it seems to me to be broken, in the sense that it does not correctly pick up Marshaler implementations that take pointer receivers when a non-pointer struct is marshaled. While at some point the yaml marshaler may want to treat pointers differently from non-pointers (e.g., to create anchors and references), this is not currently a feature so I think "casting" everything to a pointer before doing the type assertion is a valid way to go. time.Time doesn't need a custom marshaler since that type already implements TextMarshaler; and there seemed to be inconsistent handling of Duration and Time pointers vs non-pointers. This is cleaned up in the PR because the "cast" to a pointer folds. I saw other PRs that introduce the ability to set Time formatters etc but I think maybe a more general impl such as passing in a func() that gets called on all types that don't have a Marshaler so as to be able to override the default behavior ...

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -24,7 +24,6 @@ import (
 	"sort"
 	"strconv"
 	"strings"
-	"time"
 	"unicode/utf8"
 )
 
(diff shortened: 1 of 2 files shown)


aathan:
See #981 for combined PR.

PR #980, opened 2023-08-09
struct_opter

aathan:
It's a pain to declare every field in a struct as `omitempty` so this allows you to declare a function that tells the encoder to treat all fields that way. The interface can be expanded in the future to specify more "struct wide" options.


aathan:
See #981 for combined PR.

PR #981, opened 2023-08-09
Aathan combined -- all aathan PRs combined

aathan:
Since certain PRs conflict with each other this PR merges them all and resolves the conflicts. I'll mention this in all the other PRs.

PR #982, opened 2023-08-09
marshal opts

aathan:
Allows passing of options to marshal/encode ... implements one option so far which is to not lowercase all struct keys.

### DIFF
--- a/yaml.go
+++ b/yaml.go
@@ -17,8 +17,7 @@
 //
 // Source code and other details for the project are available at GitHub:
 //
-//   https://github.com/go-yaml/yaml
-//
+//	https://github.com/go-yaml/yaml
 package yaml
 
 import (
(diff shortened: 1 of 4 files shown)


aathan:
See #981 for combined PR.

yaml.go
```diff
@@ -48,7 +47,7 @@ type obsoleteUnmarshaler interface {
 // If an error is returned by MarshalYAML, the marshaling procedure stops
 // and returns with the provided error.
 type Marshaler interface {
-	MarshalYAML() (interface{}, error)
```
dolmen, line 51: This is a major breaking change.

aathan: I'm no longer working on the downstream project and do not have time to make improvements/changes to the project. I'm sure there is a way to integrate these ideas that "breaks" the existing API less, while using most of the implementation I've already provided. YAML marshaling is much more likely to need "modifiers" (options) than typical simple json marshaling, so I think you should consider adding such features. E.g. perhaps the marshaler could look for an alternative more capable interface (Marshaler2) and use it if it's available.

PR #994, opened 2023-09-26
introduce NewDecoderWithParser with ability to provide a custom parser

### LINKED ISSUE #993: how to decode a yaml.Node document ?
I'm running some custom logic to post-process a yaml document as a plain `yaml.Node`, which I obtain by :
```go
	var nodes yaml.Node
	if err := decoder.Decode(&nodes); ...
``` 
but then I have no way to invoke the actual `decoder` code. 

AFAICT [Decoder](https://github.com/go-yaml/yaml/blob/v3/yaml.go#L94) only access the `*parser` once by calling `parse()`. Makes me wonder it would sound correct to refactor this so it relies on some `Parser` interface, so that one can use:
```go
decoder := &Decoder{
		Parser: customParserProducingNodes(),
	}
decoder.Decode(&myStruct)
```

would this sound good? Happy to prepare a PR for this purpose

ndeloof:
this offers user ability to provide yaml document to be decoded into go structs as an abstract `yaml.Node` producer by encapsulating the io.Reader-based `parser` behind a `Parser` interface.

closes https://github.com/go-yaml/yaml/issues/993

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -144,7 +144,7 @@ func (p *parser) anchor(n *Node, anchor []byte) {
 	}
 }
 
-func (p *parser) parse() *Node {
+func (p *parser) Parse() *Node {
 	p.init()
 	switch p.peek() {
 	case yaml_SCALAR_EVENT:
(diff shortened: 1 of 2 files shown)


ndeloof:
@niemeyer  what do you think about this approach, does it sound right to better separate the `parser` and `decoder` into reusable components?

PR #1008, opened 2023-12-05
Add KnownFields()

fornellas:
I've been missing the ability to detect when there are alien fields when decoding. This PR exposes such interface, and add some tests.

### DIFF
--- a/node_test.go
+++ b/node_test.go
@@ -19,6 +19,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"reflect"
 
 	. "gopkg.in/check.v1"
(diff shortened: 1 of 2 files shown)


MeNsaaH:
Maybe https://github.com/kubernetes-sigs/yaml/tree/master/goyaml.v3? It's a fork by the kubernetes team and I believe it's maintained

crqrdotcom:
This would fix #460 and looks like a fairly simple implementation that aligns with the way `yaml.NewDecoder(r).KnownFields(true)` works.

Could we get this merge @niemeyer?

kurtisvg:
It would be great to see this merged!

PR #1017, opened 2024-02-07
Make yaml.Node.Decode customizable with decoder options

tonickkozlov:
Addresses #460

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -1703,6 +1703,35 @@ func (s *S) TestUnmarshalKnownFields(c *C) {
 	}
 }
 
+type decoderOptionsTest struct {
+	Foo int
+	Bar string
+}
+
+// contrived example only for the sake of showcasing the
+// custom unmarshalling logic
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,4 +1,4 @@
-module "gopkg.in/yaml.v3"
+module "github.com/tonickkozlov/yaml"
```
victor-perov: I think this is a leftover of local testing?

victor-perov (changes requested):

kvap (approved):

dolmen:
This breaks the existing API.

dolmen:
See #1008 which is an alternate implementation which doesn't break the existing API.

tonickkozlov:
Hi and thanks for your review.Could you please elaborate on how is this change breaking the API?On 21 Feb 2024, at 14:55, Olivier Mengué ***@***.***> wrote:﻿
This breaks the existing API.

PR #1018, opened 2024-02-08
Allow for custom struct tags such as `json`.

naquin:
This is useful when using structs from 3rd-party packages which almost certainly only have `json` tags and not `yaml` tags.

This implementation does not change the default behavior for only looking for `yaml` tags, but allows the user to specify a list of tags to consider such as []string{"yaml", "json"} in an encoder or decoder.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -318,11 +318,12 @@ type decoder struct {
 	stringMapType  reflect.Type
 	generalMapType reflect.Type
 
-	knownFields bool
-	uniqueKeys  bool
-	decodeCount int
-	aliasCount  int
-	aliasDepth  int
+	knownFields   bool
+	uniqueKeys    bool
+	decodeCount   int
+	aliasCount    int
+	aliasDepth    int
+	structTagKeys []string
 
 	mergedFields map[interface{}]bool
 }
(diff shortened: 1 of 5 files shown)


dolmen:
The default naming for keys which have no struct tags at all is different between `gopkg.in/yaml.v3` and `encoding/json`: `encoding/json` use lower case while `yaml.v3` keeps the original case. Here is an example: https://go.dev/play/p/L-oDNQ3nd9G

So I think that handling `json` struct tag will be the universal solution that you expect. It will only bring different breaking changes in serialization and more headaches for developers.

zcc0077:
> The default naming for keys which have no struct tags at all is different between `gopkg.in/yaml.v3` and `encoding/json`: `encoding/json` use lower case while `yaml.v3` keeps the original case. Here is an example: https://go.dev/play/p/L-oDNQ3nd9G
> 
> So I think that handling `json` struct tag will be the universal solution that you expect. It will only bring different breaking changes in serialization and more headaches for developers.

yes, the behavior is little different, for example, `json:"Abc"` can parse "abc" in json file without error by `encoding/json` but fail in `gopkg.in/yaml.v3` (empty string as result instead)

PR #1027, opened 2024-03-12
Tidy go.mod, add go.sum

dolmen:
Tidy go.mod with "go 1.12" (the first version of Go that has the "go" directive in go.mod).
Add go.sum.

```console
go mod edit -go=1.12
go mod tidy
```

This is an alternative to #877.


dolmen:
Almost a duplicate of #488, except I'm adding the "go" directive.

janisz (approved):

PR #1031, opened 2024-04-14
Fix simple key handling

### LINKED ISSUE #849: Key size in dicts seems to be limited to 128 characters
Hello,

Originally, I raised an issue againt [yq](https://github.com/mikefarah/yq): [ Strange formatting when the key size is greater than 128 characters #1192](https://github.com/mikefarah/yq/issues/1192)

But, since yq is using go-yaml v3, I was pointed here.

Maybe someone here can explain the key size limit or remove it?

Thanks and best regards

DMaxter:
Currently if a key is more than 128 characters, it is handled as a complex mapping key, however this goes against the YAML spec as pointed out in https://github.com/go-yaml/yaml/issues/849#issuecomment-1215231786 . Can be read here: https://yaml.org/spec/1.2.2/#flow-mappings

This causes issues when handling OpenAPI specifications with paths longer than 128 characters, making them invalid.

This PR aims to fix this issue, which is also affecting yq as can be seen here https://github.com/mikefarah/yq/issues/1192

Fixes #849

### DIFF
--- a/emitterc.go
+++ b/emitterc.go
@@ -997,7 +997,7 @@ func yaml_emitter_check_simple_key(emitter *yaml_emitter_t) bool {
 	default:
 		return false
 	}
-	return length <= 128
+	return length <= 1024
 }
 
 // Determine an acceptable scalar style.


AntoineJac:
Does anyone know if this PR will accepted and the release version for this feature? Thanks

DMaxter:
@AntoineJac I contacted the maintainer via email but got no answer. This project seems completely abandoned. But worth tagging @niemeyer , maybe we are lucky

niemeyer:
Sorry, I'm still here. I just haven't had the time to provide feedback and work on the project that is due.  As I said elsewhere, I want to get a team to be responsible for this so it's not a one man show, but I've been failing to allocate that.

As far as this PR is concerned, it seems reasonable, but it needs:

1) Assessment regarding backwards compatibility
2) Tests
3) Some comments about whether unicode is being handled correctly (the spec talks about unicode, not bytes)

Tieske:
tried to fix this in #1037

DMaxter:
Thank you @Tieske I think it is sensible to close this PR in favor of yours, since yours is more complete!

PR #1039, opened 2024-07-11
fix: decode !!binary as []byte

setrofim:
When decoding !!binary do not assume that the result will be a string, as !!binary can be use to encode arbitrary data.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -633,13 +633,13 @@ var unmarshalTests = []struct {
 	// Binary data.
 	{
 		"a: !!binary gIGC\n",
-		map[string]string{"a": "\x80\x81\x82"},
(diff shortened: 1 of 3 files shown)


decode_test.go
```diff
@@ -26,7 +26,7 @@ import (
 	"time"
 
 	. "gopkg.in/check.v1"
-	"gopkg.in/yaml.v3"
+	"github.com/setrofim/yaml"
```
dolmen: Do not change the package name.

dolmen (changes requested):

decode_test.go
setrofim: Ah, my apologies, that commit was not supposed to be added to the pull. Now reverted.

PR #1052, opened 2024-10-03
Fix encoding for explicitly marked str nodes

torkve:
When yaml.v3 encodes string nodes, that are explicitly marked with tag "!!str", it can remove this tag. But with ambiguous values like "y", "yes", "no" etc. it can lead to the result YAML document to be misinterpreted by YAML 1.1 compatible parsers.
This patch enforces old-boolean-like string nodes to be forcedly quoted.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -447,11 +447,13 @@ func (e *encoder) node(node *Node, tail string) {
 				tag = ""
 			} else {
 				rtag, _ := resolve("", node.Value)
-				if rtag == stag {
+				if stag == strTag {
 					tag = ""
-				} else if stag == strTag {
+					if rtag != stag || isBase60Float(node.Value) || isOldBool(node.Value) {
+						forceQuoting = true
+					}
+				} else if stag == rtag {
 					tag = ""
-					forceQuoting = true
 				}
 			}
 		} else {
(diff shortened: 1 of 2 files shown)


torkve:
Real world example: Google Gnostic compiler for OpenAPI [encodes](https://github.com/google/gnostic-models/blob/577692584cdf09674595bba0c182d0e8a44655d3/compiler/helpers.go#L274) any single OpenAPI property name as YAML string, and if one have some Point-like model `Point(x int, y int)`, after compilation it would become:
```go
&yaml.Node{
	Kind: yamlv3.MappingNode,
	Content: []*yamlv3.Node{
		{Kind:  yamlv3.ScalarNode, Tag: "!!str", Value: "x"},
		{Kind:  yamlv3.ScalarNode, Tag: "!!int", Value: "2"},
		{Kind:  yamlv3.ScalarNode, Tag: "!!str", Value: "y"},
		{Kind:  yamlv3.ScalarNode, Tag: "!!int", Value: "3"},
	},
}
```

And after serialization it becomes
```yaml
point:
  x: 2
  y: 3
``` 

Which would be deserialized incorrectly by YAML 1.1 compatible parser (e.g. one in kubernetes), since `y` would become `true`.

To the best of my knowledge I believe that Gnostic is correct here, since it explicitly marks node as string one and expects that it would always be a string in YAML, not boolean. That's why I propose this PR as a fix.

